PR #4, opened 2020-11-16
Add GetServiceEndpoint method

utkarsh-pro:
**Description**

This PR adds:
1. CreateFile utility function
2. GetServiceEndpoint method for finding endpoints for the given service in a namespace

**Notes for Reviewers**
As discussed with @kumarabd, the current approach first looks for a loadbalancer endpoint, fallsback to nodeport and then to clusterip endpoint. However, we can look at `spec.type` which does appear to indicate the type of the endpoint, can we use that instead of falling back?

### DIFF
--- a/errors/codes.go
+++ b/errors/codes.go
@@ -35,15 +35,16 @@ var (
 
 	// Meshkit specific codes
 	// Range 10200 to 10299
-	ErrSmiInit       = "kit_10200"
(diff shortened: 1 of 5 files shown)


kumarabd (approved):
LGTM

leecalcote:
Looking good @utkarsh-pro

PR #5, opened 2020-11-22
added controller protobuf

kumarabd:

**Description**

This PR fixes #

**Notes for Reviewers**


protobuf/common/error.proto
```diff
@@ -0,0 +1,14 @@
+syntax = "proto3";
+
+package common;
+
+option go_package = "github.com/layer5io/meshkit/protobuf/common;common";
+
+message CommonError {
```
mgfeller, line 7: Just `Error`? It's `common` as it is in a package called `common`, isn't it?
Will there be other types of errors, i.e. not `Common`, here in the `common` package?

protobuf/controller/health.proto
```diff
@@ -0,0 +1,14 @@
+syntax = "proto3";
+
+package controller;
+
+option go_package = "github.com/layer5io/meshkit/protobuf/controller;controller";
+
+enum ControllerStatus {
```
mgfeller, line 7: Is it common to prepend entities with package name?

protobuf/controller/health.proto
```diff
@@ -0,0 +1,14 @@
...
+	RUNNING = 0;
+	IDLE = 1;
```
mgfeller, line 9: unfamiliar with running vs idle - never came across idle as status in software....

mgfeller (commented):
@kumarabd could you please point me to some design doc or similar about the controller?

kumarabd:
> @kumarabd could you please point me to some design doc or similar about the controller?

Sure yes @mgfeller - https://docs.google.com/document/d/1AMi5UTUwLRFeQRQj7YuRQgMG-4IXWie24z4R88Eoazs/edit

protobuf/common/error.proto
```diff
@@ -0,0 +1,14 @@
...
+    string code = 1;
+    string severity = 2;
+    string shortDescription = 3;
```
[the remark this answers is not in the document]
kumarabd: @leecalcote Same question what I had im mind

protobuf/common/error.proto
```diff
@@ -0,0 +1,14 @@
...
+    string longDescription = 4;
```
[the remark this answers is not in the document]
kumarabd: @leecalcote Same question what I had im mind

protobuf/common/error.proto
kumarabd: The redundancy is mandatory here as if we use just `Error` in common package then you cannot use `Error` in any package that imports this package. @mgfeller

protobuf/controller/health.proto
kumarabd: Yes same as explained above, if we use `Status` here then any package that imports this one cannot have `Status` defined anywhere.

protobuf/controller/health.proto
kumarabd: So the use case is to get the status of the pipeline defined in the controller, @mgfeller you can suggest any better language here that you think would be good to use :)

protobuf/common/error.proto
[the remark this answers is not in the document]
kumarabd: Pulling this in discussion to the doc and going ahead with the existing way.

PR #15, opened 2020-11-27
remove default namespace and namespace overrides

utkarsh-pro:
Signed-off-by: Utkarsh Srivastava <srivastavautkarsh8097@gmail.com>

**Description**
1. This PR removes the namespace overrides. Now it falls back to namespace specified in the manifest if the ApplyOptions has no custom namespace specified. 
2. The ApplyManifest function no longer fails when namespace creation fails due to invalid namespace name (empty string), which may happen for api-resources which are not namespaced.

### DIFF
--- a/utils/kubernetes/apply-manifest.go
+++ b/utils/kubernetes/apply-manifest.go
@@ -2,7 +2,6 @@ package kubernetes
 
 import (
 	"context"
-	"reflect"
 	"strings"
 
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


mgfeller:
Great! 

Can't quite reconcile `This PR removes the namespace overrides.` (in this PR's description) and `// The namespace specified in ApplyOptions is used` doc in the code (great that the doc got updated!!)

As an aside, I wondered why length is compared to  `1` instead of `0` here:
```
if err == nil && len(val) > 1 {
			if len(options.Namespace) > 1
```

mgfeller (approved):

mgfeller:
@utkarsh-pro @kumarabd just wondering: are you testing custom operation as well?

kumarabd:
> Great!
> 
> Can't quite reconcile `This PR removes the namespace overrides.` (in this PR's description) and `// The namespace specified in ApplyOptions is used` doc in the code (great that the doc got updated!!)
> 
> As an aside, I wondered why length is compared to `1` instead of `0` here:
> 
> ```
> if err == nil && len(val) > 1 {
> 			if len(options.Namespace) > 1
> ```

Just to account for whitespaces because we don't strip it before removing them @mgfeller

kumarabd:
> @utkarsh-pro @kumarabd just wondering: are you testing custom operation as well?

I didn't, may be @utkarsh-pro did as he said he's testing out almost everything.

kumarabd (approved):

leecalcote:
I think that we should strip leading and trailing whitespace in the text fields, so that any ambiguity is removed.

leecalcote:
Leading and trailing whitespace should also be removed here in MeshKit. Will you confirm that we are stripping leading and trailing whitespace?

mgfeller:
> > Great!
> > Can't quite reconcile `This PR removes the namespace overrides.` (in this PR's description) and `// The namespace specified in ApplyOptions is used` doc in the code (great that the doc got updated!!)
> > As an aside, I wondered why length is compared to `1` instead of `0` here:
> > ```
> > if err == nil && len(val) > 1 {
> > 			if len(options.Namespace) > 1
> > ```
> 
> Just to account for whitespaces because we don't strip it before removing them @mgfeller

Ok.... exactly 1 whitespace, where is that coming from? I agree that leading and trailing whitespaces should always be removed.

PR #22, opened 2020-12-17
added logger and manifest features

kumarabd:

**Description**
1. Added logger support for controller logging interface.
2. Split ApplyManifest() to fetch runtime.Object separately.
3. Another member function to return unstructured.Unstructured{} for a given manifest.

I think this is ready for a minor version upgrade. @mgfeller @utkarsh-pro your thoughts?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ replace (
 
 require (
 	github.com/go-kit/kit v0.10.0
+	github.com/go-logr/logr v0.1.0
 	github.com/golang/protobuf v1.4.2
 	github.com/google/uuid v1.1.1
 	github.com/kr/pretty v0.2.1 // indirect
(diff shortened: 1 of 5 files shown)


mgfeller:
@kumarabd not sure about the purpose of the controller logger and where it is used, some documentation could help.
Minor version upgrade do you mean to 0.2.0 (as per semver?). It's not a bugfix, but adds functionality (Controller Logger) in a backward compatible way?

kumarabd:
@mgfeller https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.7.0/pkg/log#SetLogger for your reference on the controller-runtime logger interface that it uses. This is needed for the custom controller manager (Meshery operator) that we are working on currently.

Yes 0.2.0, yes all of them backward compatible as we are not changing the interface. Rather adding an additional method. Does it make sense?

mgfeller:
@kumarabd thanks. yes, makes sense.

PR #45, opened 2021-04-06
service discovery tcp check

kumarabd:

**Description**

1. Added TCP checks to ensure the endpoint is reachable.
2. Port discovery without port-selector returns random port from the sequence.

**Notes for Reviewers**

### DIFF
--- a/utils/kubernetes/error.go
+++ b/utils/kubernetes/error.go
@@ -2,25 +2,31 @@ package kubernetes
 
 import "github.com/layer5io/meshkit/errors"
 
-const (
-	ErrNewKubeClientCode = "test_code"
-	ErrNewDynClientCode  = "test_code"
-	ErrNewDiscoveryCode  = "test_code"
-	ErrNewInformerCode   = "test_code"
+var (
+	ErrApplyManifestCode    = "test_code"
+	ErrServiceDiscoveryCode = "test_code"
+	ErrApplyHelmChartCode   = "test_code"
+	ErrNewKubeClientCode    = "test_code"
+	ErrNewDynClientCode     = "test_code"
+	ErrNewDiscoveryCode     = "test_code"
+	ErrNewInformerCode      = "test_code"
+	ErrEndpointNotFoundCode = "test_code"
+
+	ErrEndpointNotFound = errors.NewDefault(ErrEndpointNotFoundCode, "Unable to discover an endpoint")
 )
 
 func ErrApplyManifest(err error) error {
-	return errors.NewDefault(errors.ErrApplyManifest, "Error Applying manifest", err.Error())
+	return errors.NewDefault(ErrApplyManifestCode, "Error Applying manifest", err.Error())
 }
 
 // ErrServiceDiscovery returns an error of type "ErrServiceDiscovery" along with the passed error
 func ErrServiceDiscovery(err error) error {
-	return errors.NewDefault(errors.ErrServiceDiscovery, "Error Discovering service", err.Error())
+	return errors.NewDefault(ErrServiceDiscoveryCode, "Error Discovering service", err.Error())
 }
 
 // ErrApplyHelmChart is the error which occurs in the process of applying helm chart
 func ErrApplyHelmChart(err error) error {
-	return errors.NewDefault(errors.ErrApplyHelmChart, "Error applying helm chart", err.Error())
+	return errors.NewDefault(ErrApplyHelmChartCode, "Error applying helm chart", err.Error())
 }
 
 // ErrApplyHelmChart is the error which occurs in the process of applying helm chart
--- a/utils/kubernetes/error.go
+++ b/utils/kubernetes/error.go
@@ -11,8 +11,10 @@ var (
 	ErrNewDiscoveryCode     = "test_code"
 	ErrNewInformerCode      = "test_code"
 	ErrEndpointNotFoundCode = "test_code"
+	ErrInvalidAPIServerCode = "test_code"
 
 	ErrEndpointNotFound = errors.NewDefault(ErrEndpointNotFoundCode, "Unable to discover an endpoint")
+	ErrInvalidAPIServer = errors.NewDefault(ErrInvalidAPIServerCode, "Invalid API Server URL")
 )
 
 func ErrApplyManifest(err error) error {
--- a/utils/kubernetes/service.go
+++ b/utils/kubernetes/service.go
@@ -33,12 +33,14 @@ func GetEndpoint(ctx context.Context, opts *ServiceOptions, obj *corev1.Service)
 	var nodePort, clusterPort int32
 	endpoint := utils.Endpoint{}
 
-	if opts.PortSelector != "" {
-		for _, port := range obj.Spec.Ports {
-			if port.Name == opts.PortSelector {
-				nodePort = port.NodePort
-				clusterPort = port.Port
-			}
+	for _, port := range obj.Spec.Ports {
+		if opts.PortSelector != "" && port.Name == opts.PortSelector {
+			nodePort = port.NodePort
+			clusterPort = port.Port
+			break
+		} else {
+			nodePort = port.NodePort
+			clusterPort = port.Port
 		}
 	}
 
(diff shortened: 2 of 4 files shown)


utils/kubernetes/service.go
```diff
@@ -33,12 +33,14 @@ func GetEndpoint(ctx context.Context, opts *ServiceOptions, obj *corev1.Service)
...
 
-	if opts.PortSelector != "" {
-		for _, port := range obj.Spec.Ports {
-			if port.Name == opts.PortSelector {
-				nodePort = port.NodePort
-				clusterPort = port.Port
-			}
+	for _, port := range obj.Spec.Ports {
+		if opts.PortSelector != "" && port.Name == opts.PortSelector {
+			nodePort = port.NodePort
```
mgfeller, line 38: not quite sure I follow the logic here, but if the desired port is the second (or third...) in the list, and `opts.PortSelector != ""` does it actually reach it?

utils/kubernetes/service.go
```diff
@@ -73,5 +75,20 @@ func GetEndpoint(ctx context.Context, opts *ServiceOptions, obj *corev1.Service)
 		}
 	}
 
+	if !utils.TcpCheck(endpoint.External) {
+		if len(strings.SplitAfter(opts.APIServerURL, "://")) > 1 {
```
mgfeller, line 79: I find it a bit difficult to understand the logic here, as I don't know enough about why `APIServerURL` is used in some cases (something specific to `minikube`). It would really be helpful to have an example of a `APIServerURL` here that would result in a valid endpoint.

utils/kubernetes/service.go
```diff
@@ -73,5 +75,20 @@ func GetEndpoint(ctx context.Context, opts *ServiceOptions, obj *corev1.Service)
...
+			endpoint.External.Address = strings.SplitAfter(strings.SplitAfter(opts.APIServerURL, "://")[1], ":")[0]
+			if len(endpoint.External.Address) > 0 {
+				endpoint.External.Address = endpoint.External.Address[:len(endpoint.External.Address)-1]
+			}
+		}
+		if !utils.TcpCheck(endpoint.External) {
```
mgfeller, line 85: check whether looking at `APIServerURL` return something alive...

utils/kubernetes/service.go
```diff
@@ -73,5 +75,20 @@ func GetEndpoint(ctx context.Context, opts *ServiceOptions, obj *corev1.Service)
...
+			endpoint.External.Port = nodePort
```
mgfeller, line 86: ... nope, try the `NodePort`

utils/network.go
```diff
@@ -18,9 +18,9 @@ type HostPort struct {
 	Port    int32
 }
 
-func TcpCheck(ip string, port int32) bool {
+func TcpCheck(hp *HostPort) bool {
 	timeout := 5 * time.Second
-	conn, err := net.DialTimeout("tcp", fmt.Sprintf("%s:%d", ip, port), timeout)
+	conn, err := net.DialTimeout("tcp", fmt.Sprintf("%s:%d", hp.Address, hp.Port), timeout)
```
mgfeller, line 23: can `conn != nil` even if `err !=nil`? if not one could use:
```
        conn, _ := net.DialTimeout("tcp", fmt.Sprintf("%s:%d", hp.Address, hp.Port), timeout)
	if conn != nil {
		return true
	}
	return false
```

utils/kubernetes/service.go
kumarabd: The PortSelector would be the name of the port that we would want to connect to, so it would point to the exact item in the list where the `Name` matches the selector. The desired port cannot be based on index, because the order can be varying. So if no PortSelector is specified, it would return the last `Port` element in the list (Which works out well if there is only one port to connect to) @mgfeller

utils/kubernetes/service.go
kumarabd: So in case of Minikube hyper kit clusters, the LoadBalancer IP turns out to be the same as that of the ClusterIP. So in this scenario it means that the LoadBalancer IP is nothing but the APIServer's IP.So here we strip off the IP and check if that's reachable, does this make sense? @mgfeller

utils/kubernetes/service.go
kumarabd: It wouldn't matter correct? As long as the port responds, correct me if I'm wrong. @mgfeller

utils/kubernetes/service.go
kumarabd: U mean `port.NodePort`? If yes, this is exactly the same as that.

utils/network.go
kumarabd: conn cannot be non-nil when err is non-nil, so this works out every time.

utils/kubernetes/service.go
mgfeller: no, it wouldn't, I was just "thinking aloud" :smile:

utils/kubernetes/service.go
mgfeller: Ah yes.

utils/network.go
mgfeller: I always have troubles with double negations :smile: error non-nil means conn = nil, right? So the code here can be simplified?
Can't see that this is explained here either https://golang.org/pkg/net/#Dial.

utils/kubernetes/error.go
```diff
@@ -2,25 +2,31 @@ package kubernetes
...
-	ErrNewInformerCode   = "test_code"
+var (
+	ErrApplyManifestCode    = "test_code"
+	ErrServiceDiscoveryCode = "test_code"
+	ErrApplyHelmChartCode   = "test_code"
+	ErrNewKubeClientCode    = "test_code"
+	ErrNewDynClientCode     = "test_code"
+	ErrNewDiscoveryCode     = "test_code"
+	ErrNewInformerCode      = "test_code"
+	ErrEndpointNotFoundCode = "test_code"
```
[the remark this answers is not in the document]
kumarabd: This change is not related to the error code. Will be addressing it in upcoming PRs. @leecalcote

utils/kubernetes/error.go
```diff
@@ -2,25 +2,31 @@ package kubernetes
...
+
+	ErrEndpointNotFound = errors.NewDefault(ErrEndpointNotFoundCode, "Unable to discover an endpoint")
```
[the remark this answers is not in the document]
kumarabd: This change is not related to the error code. Will be addressing it in upcoming PRs. @leecalcote

utils/kubernetes/service.go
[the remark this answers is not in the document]
kumarabd: yes I can replace this with url library

utils/kubernetes/error.go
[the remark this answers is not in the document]
mgfeller: @leecalcote, @kumarabd mentioned that  `errors.NewDefault` should be replaced by `errors.New` (I understand this to  mean deprecating/removing `errors.NewDefault`). Using  `errors.New` means that one is strongly reminded to add probable cause and suggested remediation.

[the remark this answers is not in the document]
kumarabd: Yes @mgfeller , basically that should be of a whole different pull request I believe.

utils/network.go
kumarabd: yes I think that should work. Ill simplify the condition. Thanks @mgfeller

utils/kubernetes/service.go
kumarabd: Ill work on that, yes sounds good. @mgfeller

kumarabd:
@mgfeller Added test cases, does this do the trick?

utils/kubernetes/error.go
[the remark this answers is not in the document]
mgfeller: @kumarabd yes I agree.

leecalcote (approved):

mgfeller (approved):

PR #63, opened 2021-06-19
 [Chore] Relocate errorutils files from root directory

### LINKED ISSUE #62: [Chore] Relocate errorutils files from root directory
**Note**: original issue in Meshery repository https://github.com/layer5io/meshery/issues/3207
See also discussion there.

#### Current Behavior
Four new errorutils files have landed in the root directory of the Meshery repo (e.g. https://github.com/layer5io/meshery/blob/master/errorutil_errors_export.json, https://github.com/layer5io/meshery/blob/master/component_info.json).

#### Desired Behavior
Ideally, these files are located in a separate folder, like "helpers".

#### Implementation
Move the errorutils files into a directory (e.g. "helpers/errorutils/")

mgfeller:
Introduced 2 new command line params: output directory (`-o`) and directory for `component_info.json` (`-i`)
Default values are the root directory (`-d`).

some refactoring to remove code duplication etc

other changes:

- int_codes are now sorted for better comparison between runs
- (re)added make target to build errorutil binary
- only error.go files are updated
- only files actually updated by errorutil are written back to disc
- default log level set to WARN
- regex for error codes changed to Err[A-Z].+Code

Signed-off-by: Michael Gfeller <mgfeller@mgfeller.net>

**Description**

This PR fixes #62 

**Notes for Reviewers**


mgfeller:
@DelusionalOptimist @kumarabd @Chadha93 @leecalcote based on yesterday's feedback I have updated this PR with some more changes: 

- only error.go files are updated
- only files actually updated by errorutil are written back to disc
- default log level set to WARN
- regex for error codes changed to Err[A-Z].+Code

Makefile
```diff
@@ -7,8 +7,14 @@ check-clean-cache: error
...
 
 errorutil:
-	go run github.com/layer5io/meshkit/cmd/errorutil -d . update
\ No newline at end of file
+	go run github.com/layer5io/meshkit/cmd/errorutil -d . update
+
+errorutil-analyze:
+	go run github.com/layer5io/meshkit/cmd/errorutil -d . analyze
+
+build-errorutil:
```
kumarabd, line 19: @mgfeller I think we can get rid of line 10:11 because we have line 13:14. What do you think?

kumarabd:
Rest LGTM

Makefile
mgfeller: @kumarabd not quite sure, the purpose is to run the locally built binary for testing purposes just in this repo when developing the utility, in a branch. I assume `go run github.com/layer5io/meshkit/cmd/errorutil` reflects the latest remote version. So I think for that use case, `errorutil` is not equivalent to `error`. But it could be removed anyway, if you want.

kumarabd: No problem, sure keep it doesn't harm anything. Lets go with this PR.

PR #80, opened 2021-07-09
[ErrorCodes] errorutil output files in sync with codes updated by the utility

### LINKED ISSUE #74: [ErrorCodes] errorutil output files should be in sync with updated code
**Current Behavior**

* The output files do not reflect the updated code, but the previous state.
*  `next_error_code` is not updated if codes are changed manually

**Desired Behavior**

* The output files should reflect the state of the code after error codes have been updated.
* an error should be issued if `next_error_code` is not greater than the highest code in the code (later, set exit status 1, see #65). 

**Workaround**

* Run `updated` twice - `errorutil_*` are in sync-
* Update `next_error_code` manually

**Testing**

* no int codes are set, all with string such as "replace_me", running update should produce correct output files, e.g. no duplicates ("replace_me"), repeatedly running `update` or `analyze` (which also generates the output files) should not change output files

* some codes are set, others with string such as "replace_me", running update should produce correct output files, e.g. no duplicates ("replace_me"), repeatedly running `update` or `analyze` (which also generates the output files) should not change output files

* set `next_error_code`  to a number smaller than the highest code, should produce an error

---
**Resources**

**Alternatives / Additional Context**

mgfeller:
**Description**

This PR fixes #74 

**Notes for Reviewers**

* In case of `update`, an analysis/validation is performed after the update, (the equivalent of running `analyze`), meaning that there are 2 passes through the code.
* The output files reflect now the state of the code after error codes have been updated.
* An error is be issued if `next_error_code` is not greater than the highest code in the code.


leecalcote (approved):

sayantan1413 (approved):
Hey @mgfeller This looks good to me. But I have question like from now on all the three JSON files and the component_info.json with go under the `helpers` directory and that means that we have update all the makefile to `go run github.com/layer5io/meshkit/cmd/errorutil -d . update -i ./helpers -o ./helpers`

mgfeller:
> Hey @mgfeller This looks good to me. But I have question like from now on all the three JSON files and the component_info.json with go under the `helpers` directory and that means that we have update all the makefile to `go run github.com/layer5io/meshkit/cmd/errorutil -d . update -i ./helpers -o ./helpers`

Thanks @sayantan1413 to answer your question: it appears so, I didn't take part in that discussion / decision.

PR #82, opened 2021-07-09
fix: Combined same types of function params

iamrajiv:
## Description

If parameters of the same type lie consecutively, mention their type once at the end of the last parameter. Unlike other languages, like C, where all parameters must be specified with types explicitly, it is not required to do so in Go. Combining the types is usually recommended for the sake of brevity.

### DIFF
--- a/cmd/errorutil/internal/coder/commands.go
+++ b/cmd/errorutil/internal/coder/commands.go
@@ -64,7 +64,7 @@ func getGlobalFlags(cmd *cobra.Command) (globalFlags, error) {
 	return flags, nil
 }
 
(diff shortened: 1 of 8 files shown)


mgfeller (commented):
Thanks @iamrajiv . I think this is also a matter of taste. I prefer explicit types after each parameters, it means I have to parse and think less (i.e. "don't make me think" coding principle :smile: ). If multiple parameters of same type have just one type declaration, I'd prefer that they are on separate lines, again making it easier to parse and understand. IMHO brevity is not always preferable, code readability is.

Issue #13, opened 2020-11-25
Add golangci-lint config, fix reported errors, add gtihub workflow


mgfeller:
**Current Behavior**

Currently, golangci-lint is not used to test code quality.

**Desired Behavior**

* Use golangci-lint to test code quality.
* Fix reported errors
* Add .github workflow ci.yml

Config and ci.yml can be copied from other repo, e.g. meshery-consul.

PR #86, opened 2021-07-19
feat[CI]: Add golangci-lint

iamrajiv:
Fixes: #13

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,0 +1,21 @@
+name: golangci-lint
+on:
+  push:
+    branches:
+      - '**'
+    tags:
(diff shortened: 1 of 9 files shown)


kumarabd:
@iamrajiv Some ci issues you want to address

mgfeller (changes requested):
@iamrajiv thanks! still the same 3 files where the imports needs to be rearranged. Not sure whether VSCode does that, in case your using it (I'm not), it can also be fixed using `goimports`, e.g. `goimports -w cmd/errorutil/internal/coder/node.go`

mgfeller (approved):
Many thanks @iamrajiv good to see this one resolved :tada:

Issue #88, opened 2021-07-20
Adding `golangci-lint` configuration


iamrajiv:
## References:

https://golangci-lint.run/usage/configuration/

mgfeller:
Some configs from other repos: 

* https://github.com/meshery/meshery/blob/master/.golangci.yml
* https://github.com/meshery/meshery-consul/blob/master/.golangci.yml
* https://github.com/meshery/meshery-operator/blob/master/.golangci.yml

iamrajiv:
@mgfeller Please specify the directories which has to be skipped in this configuration while running the `golangci-lint`.

mgfeller:
> @mgfeller Please specify the directories which has to be skipped in this configuration while running the `golangci-lint`.

`.github` I guess, possibly `helpers` (unless we suddenly get code there, not unlikely), otherwise I think all should be included.

PR #89, opened 2021-07-20
feat: Add `golangci-lint` configuration

iamrajiv:
Fixes: #88

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,2 +1,147 @@
+linters-settings:
+  depguard:
+    list-type: blacklist
+    packages:
+      # logging is allowed only by logutils.Log, logrus
+      # is allowed to use only in logutils package
+      - github.com/sirupsen/logrus
+    packages-with-error-message:
+      - github.com/sirupsen/logrus: "logging is allowed only by logutils.Log"
+  dupl:
+    threshold: 100
+  exhaustive:
+    default-signifies-exhaustive: false
+  funlen:
+    lines: 100
+    statements: 50
+  gci:
+    local-prefixes: github.com/golangci/golangci-lint
+  goconst:
+    min-len: 2
+    min-occurrences: 2
+  gocritic:
+    enabled-tags:
+      - diagnostic
+      - experimental
+      - opinionated
+      - performance
+      - style
+    disabled-checks:
+      - dupImport # https://github.com/go-critic/go-critic/issues/845
+      - ifElseChain
+      - octalLiteral
+      - whyNoLint
+      - wrapperFunc
+  gocyclo:
+    min-complexity: 15
+  goimports:
+    local-prefixes: github.com/golangci/golangci-lint
+  golint:
+    min-confidence: 0
+  gomnd:
+    settings:
+      mnd:
+        # don't include the "operation" and "assign"
+        checks: argument,case,condition,return
+  gosec:
+    settings:
+      exclude: -G204
+  govet:
+    check-shadowing: false
+    settings:
+      printf:
+        funcs:
+          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Infof
+          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Warnf
+          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Errorf
+          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Fatalf
+  lll:
+    line-length: 950
+  maligned:
+    suggest-new: true
+  misspell:
+    locale: US
+  nolintlint:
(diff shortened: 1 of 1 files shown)


mgfeller:
@iamrajiv great! the version of `golangci-lint` needs to be upgraded to support the `revive` linter - good to upgrade anyway.

logger/types.go
```diff
@@ -1,7 +1,7 @@
 package logger
 
 const (
-	JsonLogFormat = iota
+	JSONLogFormat = iota
```
mgfeller, line 4: This is a breaking change. Not sure we should change that. @kumarabd ?
It can be exclude from the linter by using a comment like `//nolint:<check>`.

utils/network.go
```diff
@@ -22,7 +22,7 @@ type MockOptions struct {
 	DesiredEndpoint string
 }
 
-func TcpCheck(hp *HostPort, mock *MockOptions) bool {
+func TCPCheck(hp *HostPort, mock *MockOptions) bool {
```
mgfeller, line 25: This is a breaking change. Not sure we should change that. @kumarabd ?
It can be exclude from the linter by using a comment like `//nolint:<check>`.

mgfeller (changes requested):
Added some comments, please have a look @iamrajiv

logger/types.go
kumarabd: I really want to change this, even if this means I replicate this across places and update.😅 But you're right, better to comment out the lint check.

mgfeller (changes requested):
@iamrajiv great! The action is failing because the version of  `golangci-lint` is too old, upgrading to 1.14.1 will solve this (https://github.com/layer5io/meshkit/blob/5bc1f9c13e60c7674d8b8c1d8ea434cbb2ce08fd/.github/workflows/golangci-lint.yml#L20)

.golangci.yml
```diff
@@ -1,2 +1,159 @@
...
+  maligned:
+    suggest-new: true
+  misspell:
+    locale: US
+  nolintlint:
+    allow-leading-space: true # don't require machine-readable nolint directives (i.e. with no leading space)
+    allow-unused: false # report any unused nolint directives
+    require-explanation: false # don't require an explanation for nolint directives
+    require-specific: false # don't require nolint directives to be specific about which linter is being skipped
+revive:
```
mgfeller, line 67: I think this need to be indented, it's a subsection of `linters-settings` (https://revive.run/docs#golangci-lint)

mgfeller (changes requested):
@iamrajiv great! next would be to get rid of the errors. 

Annotations in the code should be used to ignore the following errors (unless @kumarabd changes his mind)
```
  Error: ST1003: const JsonLogFormat should be JSONLogFormat (stylecheck)
  Error: ST1003: func TcpCheck should be TCPCheck (stylecheck)
```

As `revive` provides more rules compared to `golint`, other linters might become superfluous.
If I remember correctly from my testing yesterday, the above errors are covered both by `revive` and `stylecheck`.

Any other config changes you suggest?

.golangci.yml
```diff
@@ -1,2 +1,159 @@
...
+    ignore-generated-header: true
+    severity: warning
+    rules:
+      - name: indent-error-flow
+        severity: warning
+      - name: add-constant
+        severity: warning
+        arguments:
+          - maxLitCount: "3"
+            allowStrs: '""'
```
mgfeller, line 78: maybe add "." for current directory?

.golangci.yml
```diff
@@ -1,2 +1,159 @@
...
+            allowInts: "0,1,2"
```
mgfeller, line 79: rationale why these ints and floats should be allowed? "0" can be ok as length comparison, the others?

mgfeller:
Hi @iamrajiv, how's it going? I pushed some changes to fix linter errors and warnings in `cmd/errorutil`. Some of them were magic numbers in tests which I configured revive to ignore using annotation in the code. I think in general it is ok to have magic numbers as test data in tests (`_test.go` files). There is an exclusion rule for this in `.golangci.yml` for the `gomnd` linter, in addition also `revive` reports this, so here is an overlap too. Adding a exclusion config for this particular revive rule would be good (I just assume it is possible).

.golangci.yml
```diff
@@ -140,6 +140,7 @@ issues:
     - path: _test\.go
       linters:
         - gomnd
+        - revive
```
mgfeller, line 143: @iamrajiv hmm I don't think that _all_ `revive` rules should be disabled for test files, just selected ones, like the one for magic numbers (maybe others too). test code should in general be treated like prod code and be of the same quality.

mgfeller (changes requested):
@iamrajiv I think exclusion rules (for test files) should be more fine grained.

PR #97, opened 2021-08-06
Generating components

Revolyssup:
**Description**
Adding component generation in meshkit.

**Notes for Reviewers**


utils/kubernetes/client.go
```diff
@@ -49,21 +48,25 @@ func DetectKubeConfig(configfile []byte) (config *rest.Config, err error) {
 func processConfig(configFile []byte) ([]byte, error) {
 	cfg, err := clientcmd.Load(configFile)
 	if err != nil {
+
```
leecalcote, line 51: Bleck. Let's set some standard formatters as a convention and see if we can stick to those as a project. // @layer5io/adapter-maintainers

leecalcote: @Revolyssup in the meantime, can you revert you changes to this file?

utils/kubernetes/service_test.go
```diff
@@ -29,12 +29,12 @@ func TestGetServiceEndpoint(t *testing.T) {
 		Spec: corev1.ServiceSpec{
 			ClusterIP: "1.1.1.1",
 			Ports: []corev1.ServicePort{
-				corev1.ServicePort{
+				{
```
leecalcote, line 32: Revert changes here, please, assuming this is formatting only.

utils/manifests/definitions.go
```diff
@@ -0,0 +1,18 @@
+package manifests
+
+const template = "{\"apiVersion\":\"core.oam.dev/v1alpha1\",\"kind\":\"WorkloadDefinition\",\"metadata\":{\"name\":\"\"},\"spec\":{\"definitionRef\":{\"name\":\"\"}}}"
+const (
+	SERVICE_MESH = iota
+	K8
```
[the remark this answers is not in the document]
leecalcote: If we're going to reference service meshes by name, please use the SMP proto.

utils/manifests/generateComponent.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		return nil, err
+	}
+	var binPath string = wd + "/utils/manifests/utilBin/kubeopenapi-jsonschema"
+	switch runtime.GOOS {
+	case "windows":
+		{
+			binPath += ".exe"
+		}
+	case "darwin":
```
leecalcote, line 22: Does this mean that Linux is the default or that Linux is not supported?

leecalcote (changes requested):
I don't think that we should be bundling copies of the kubeopenapi-jsonschema tool into this repo.

utils/manifests/getComponents.go
```diff
@@ -0,0 +1,38 @@
+package manifests
+
+import (
+	"github.com/layer5io/meshkit/utils"
+	k8 "github.com/layer5io/meshkit/utils/kubernetes"
```
leecalcote, line 5: Please use "k8s"

utils/manifests/generateComponent.go
```diff
@@ -0,0 +1,74 @@
...
+		{
+			binPath += "-darwin"
+		}
+	}
+
+	//make the binary executable
+	if err = os.Chmod(binPath, 0750); err != nil {
+		return nil, err
```
leecalcote, line 30: No MeshKit-style error?

utils/manifests/generateComponent.go
Revolyssup: @leecalcote  the binary for linux is kubeopenapi-jsonschema. If the platforms are darwin or windows, these postfixes are added because that is the name of their respective binaries. And if neither darwin nor windows is the platform, then it is assumed to be linux.

utils/manifests/getComponents.go
Revolyssup: sorry, will do

utils/manifests/generateComponent.go
Revolyssup: sorry, will add.

utils/manifests/definitions.go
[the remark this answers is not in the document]
Revolyssup: I thought to use them as enums. 
sorry, will change.

Revolyssup:
Currently GetFromHelm is failing in some cases. That function has to be improved. 
Also configuration is to be enhanced so that adpaters can pass custom json path.

utils/manifests/utils.go
```diff
@@ -0,0 +1,183 @@
+package manifests
+
+import (
+	"bytes"
+	"encoding/json"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/layer5io/meshkit/models/oam/core/v1alpha1"
```
Aisuko, line 10: Shall we change the version from `v1alpha1` To `v1`?

utils/manifests/getComponents.go
```diff
@@ -0,0 +1,38 @@
+package manifests
+
+import (
+	"github.com/layer5io/meshkit/utils"
+	k8s "github.com/layer5io/meshkit/utils/kubernetes"
```
Aisuko, line 5: Why there included the binary file `utils/manifests/utilBin/kubeopenapi-jsonschema` , `utils/manifests/utilBin/kubeopenapi-jsonschema-darwin` ?

utils/manifests/definitions.go
```diff
@@ -0,0 +1,29 @@
+package manifests
+
+const template = "{\"apiVersion\":\"core.oam.dev/v1alpha1\",\"kind\":\"WorkloadDefinition\",\"metadata\":{\"name\":\"\"},\"spec\":{\"definitionRef\":{\"name\":\"\"}}}"
```
Aisuko, line 3: We  could define one constant and using it in many places, that can easy to update the code when we bump the `crd` resource version.

Aisuko (changes requested):
There are some suggestions, please take a look. And please add the test cases for all the functions, thanks.

utils/manifests/getComponents.go
leecalcote: +1. Ideally, we are not including these binaries local to the MeshKit package. ￼

utils/manifests/generateComponent.go
leecalcote: Thank you, sir.

Revolyssup:
> I don't think that we should be bundling copies of the kubeopenapi-jsonschema tool into this repo.

Where else do you suggest the binaries to be?

utils/manifests/getComponents.go
Revolyssup: @leecalcote  you told me last time to package them with binary, 
do you prefer to fetch them on the fly?

leecalcote:
> > I don't think that we should be bundling copies of the kubeopenapi-jsonschema tool into this repo.
> 
> 
> 
> Where else do you suggest the binaries to be?
> 
> 

Imported dynamically

utils/manifests/definitions.go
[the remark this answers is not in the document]
Revolyssup: > If we're going to reference service meshes by name, please use the SMP proto.

Here we are not referring to service mesh by name. The constant SERVICE_MESH is used to just check what the user passed in config, service mesh or meshery or k8s

Revolyssup:
test code
`	v := "v1.9.7"
	m := manifests.Config{
		Name:        "Istio",
		MeshVersion: v,
		Filter: manifests.CrdFilter{
			RootFilter:    []string{"$[?(@.kind==\"CustomResourceDefinition\")]"},
			NamFilter:     []string{"$..[\"spec\"][\"names\"][\"kind\"]"},
			NamePath:      "@.spec.names.kind",
			VersionFilter: []string{"$..spec.versions[0]", " --o-filter", "$[0]"},
			GroupFilter:   []string{"$..spec", " --o-filter", "$[]"},
			SpecFilter:    []string{"$..openAPIV3Schema.properties.spec", " --o-filter", "$[]"},
		},
	}
	url := "https://raw.githubusercontent.com/istio/istio/1.9.7/manifests/charts/base/crds/crd-all.gen.yaml"
	comp, err := manifests.GetFromManifest(url, manifests.SERVICE_MESH, m)
	if err != nil {
		fmt.Printf("%s", err.Error())
	}
	fmt.Println("Printing stuff: ", comp.Definitions[1])`

utils/kubernetes/get-manifests.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"helm.sh/helm/v3/pkg/action"
+	"helm.sh/helm/v3/pkg/chart/loader"
+	"helm.sh/helm/v3/pkg/cli"
+	"helm.sh/helm/v3/pkg/cli/values"
+	"helm.sh/helm/v3/pkg/getter"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+)
+
+func GetManifestsFromHelm(client *Client, url string) (string, error) {
```
DelusionalOptimist, line 12: @Revolyssup A few comments explaining what is happening here would help.
* How is this function getting manifests from helm charts?
* Is this also applying helm charts? If so, why is there a need to do so?

DelusionalOptimist: If this is applying helm charts for some reason, can this reuse https://github.com/layer5io/meshkit/blob/cbaca9a35a40eb5bc9930bbbabc6bbc2498b2fd3/utils/kubernetes/apply-helm-chart.go#L220?

utils/manifests/definitions.go
```diff
@@ -0,0 +1,30 @@
...
```
DelusionalOptimist, line 3:
```suggestion
const workloadDefinitionTemplate = "{\"apiVersion\":\"core.oam.dev/v1alpha1\",\"kind\":\"WorkloadDefinition\",\"metadata\":{\"name\":\"\"},\"spec\":{\"definitionRef\":{\"name\":\"\"}}}"
```
Let's call it this?

utils/manifests/definitions.go
```diff
@@ -0,0 +1,30 @@
...
+const (
+	SERVICE_MESH = iota
+	K8s
+	MESHERY
+)
```
DelusionalOptimist, line 8:
```suggestion
// Type of resource
const (
    // service mesh resource
	SERVICE_MESH = iota
	// native Kubernetes resource
	K8s
	// native Meshery resource
	MESHERY
)
```

utils/manifests/generateComponent.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/layer5io/meshkit/utils"
+)
+
+func generateComponents(manifest string, resource int, cfg Config) (*Component, error) {
+	wd, err := os.Getwd()
+	if err != nil {
+		return nil, err
+	}
+
+	var binPath string = wd + "/utils/manifests/utilBin/kubeopenapi-jsonschema"
```
DelusionalOptimist, line 20: A good place to store the fetched binary would be in `~/.meshery/bin` which will ensure multiple components can use it.

utils/manifests/getComponents.go
DelusionalOptimist: I think we would be fetching the binary for jsonschema utility on demand. (when the user demands to generate components).
Shipping it in either this library or the adapter image is not needed. We don't expect users to grab new service mesh manifests and generate components in air gapped deployments.

DelusionalOptimist (changes requested):

utils/kubernetes/get-manifests.go
Revolyssup: > @Revolyssup A few comments (comments in code) explaining what is happening here would help.
> 
> * How is this function getting manifests from helm charts?
> * Is this also applying helm charts? If so, why is there a need to do so?

Ans 1- I went to the source code of helm and tried to find where exactly do they populate the config with manifests. The manifests are inside the release struct which is returned by the Run function on *action.Install. So I take out the release from their and manifests are inside the release.
Ans 2- No it doesn't, I have set the "dry run" as true before executing the Run function. Although it needs to be able to communicate to kube api server(I dont know for what reason), it doesn't sends the manifests to the k8s. So no , it doesn't actually apply those helm charts.

Revolyssup: > If this is applying helm charts for some reason, can this reuse
> 
> https://github.com/layer5io/meshkit/blob/cbaca9a35a40eb5bc9930bbbabc6bbc2498b2fd3/utils/kubernetes/apply-helm-chart.go#L220
> 
> ?
Yeah I also initially thought to use this function as utkarsh has already included a dry run flag in the config but didn't use because-
This function doesn't return any release struct or manifests which we actually need. Getting logs out of here was a bit tricky for me.It seemed easier to write 30-40 lines of code. Although writing this function, I used ApplyHelmChart as reference as well as helm source code.

utils/manifests/definitions.go
Revolyssup: > Let's call it this?

Cool

utils/manifests/generateComponent.go
Revolyssup: @DelusionalOptimist  yups, didnt think of that

utils/manifests/definitions.go
```diff
@@ -0,0 +1,35 @@
+package manifests
+
+const workloadDefinitionTemplate = "{\"apiVersion\":\"core.oam.dev/v1alpha1\",\"kind\":\"WorkloadDefinition\",\"metadata\":{\"name\":\"\"},\"spec\":{\"definitionRef\":{\"name\":\"\"}}}"
```
utkarsh-pro, line 3: Please use OAM struct here.

DelusionalOptimist (approved):
Looks good to me.

utils/manifests/definitions.go
DelusionalOptimist: Addressed in https://github.com/layer5io/meshkit/pull/97/commits/123bac544df9da2fe333cfcb0fab1eda085134ed

utils/manifests/generateComponent.go
```diff
@@ -0,0 +1,97 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+
+	"github.com/layer5io/meshkit/utils"
+)
+
+func generateComponents(manifest string, resource int, cfg Config) (*Component, error) {
+	wd := utils.GetHome() + "/.meshery/bin"
```
utkarsh-pro, line 15: Windows users...

```suggestion
	wd := filepath.Join(utils.GetHome(), ".meshery", "bin")
```

utils/manifests/generateComponent.go
```diff
@@ -0,0 +1,97 @@
...
+	fmt.Println("Looking for kubeopenapi-jsonschema in ", wd)
+	var binPath string = wd + "/kubeopenapi-jsonschema"
```
utkarsh-pro, line 17: Use `filepath`

utils/manifests/generateComponent.go
```diff
@@ -0,0 +1,97 @@
...
+	//make the binary executable
+	if err := os.Chmod(binPath, 0750); err != nil {
+		return nil, err
+	}
+
+	var (
+		out bytes.Buffer
+		er  bytes.Buffer
+	)
+	path := filepath.Join(wd, "/test.yaml")
```
utkarsh-pro, line 50:
```suggestion
	path := filepath.Join(wd, "test.yaml")
```

utils/manifests/utils.go
```diff
@@ -0,0 +1,229 @@
...
+				"k8sAPIVersion": apiGroup + "/" + apiVersion,
+				"k8skind":       crd,
+			}
+		}
+	case K8s:
+		{
+			def.Spec.Metadata = map[string]string{
+				"@type":         "pattern.meshery.io/k8s",
+				"k8sAPIVersion": apiGroup + "/" + apiVersion,
+				"k8skind":       "",
```
utkarsh-pro, line 44: This is supposed to be `Kind` of resource. 
Example: `Service`, `Endpoint`, etc.

utkarsh-pro: Refer: https://github.com/meshery/meshery/blob/master/oam/workloads/service.k8s_definition.json

utils/manifests/utils.go
```diff
@@ -0,0 +1,229 @@
...
+	apiGroup, err := getApiGrp(binPath, filepath, crd)
+	if err != nil {
+		return "", ErrGetAPIGroup(err)
+	}
+	//getting defintions for different native resources
+	def.Spec.DefinitionRef.Name = definitionRef
+	def.ObjectMeta.Name = crd
+	switch resource {
+	case SERVICE_MESH:
+		{
```
utkarsh-pro, line 30: We don't need curly braces in case statements

utils/manifests/generateComponent.go
utkarsh-pro: Consider doing this in `os.TempDir()`

utkarsh-pro (changes requested):

utils/manifests/generateComponent.go
Revolyssup: That was my initial approach but empty components were returned everytime. The file was getting populated too but somehow kubeopenapi-jsonschema returns no output when file path is in /temp/test.yaml. And I have confirmed that file is there and populated. So that is why, at last I went with this. Why do you think this is happening?

utkarsh-pro (approved):
LGTM

PR #107, opened 2021-09-24
fixed bugs,enhanced filters and logic

Revolyssup:

**Description**

This PR fixes previous bugs with service mesh component generation and helps add support for k8s component generation.

**Notes for Reviewers**

### DIFF
--- a/utils/manifests/definitions.go
+++ b/utils/manifests/definitions.go
@@ -24,16 +24,44 @@ type Config struct {
 	ModifyDefSchema func(*string, *string) //takes in definition and schema, does some manipulation on them and returns the new def and schema
 }
 
+/* How to customize these filters (These comments to be updated if the behavior changes in future)-
+There are only two types of filters used internally by kubeopenapi-jsonschema which is being used here- filter(input filter) and output filter.
+The filters described below are an abstraction over those filters.
+1. RootFilter- is used at two places
+	(a) For fetching the crd names or api resources on which we will iterate over, first we apply the root filter to get the objects we are interested in and then
+		the NameFilter is applied as output filter to take out just the names from selected objects.
+	(b) [this will be discussed after ItrFilter]
+
+2. NameFilter- As explained above, it is used as an --o-filter to extract only the names after RootFilter(1(a)) has been applied.
+3. ItrFilter- This is an incomplete filter, intentionally left incomplete. Before getting version and group with VersionFilter and GroupFilter, we want to obtain only the
+			object we are interested in, in a given iteration. Crdnames or ApiResource names which are obtained by NameFilter are iterated over and used within,lets call it X.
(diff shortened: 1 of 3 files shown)


utils/manifests/definitions.go
```diff
@@ -30,5 +30,10 @@ type CrdFilter struct {
 	GroupFilter   JsonPath //This will specify the path to get to group name
 	VersionFilter JsonPath //This will specify the path to get to version name. [Version should have a name field]
 	SpecFilter    JsonPath //This will specify the path to get spec
-	IsJson        bool     //Set to true if input format is json
+	ItrFilter     string   //Filter is appended with "=="<some-crd-name>")]", so no need to write it completely. Just write the part before double equals
```
leecalcote, line 33: What does Itr stand for?

utils/manifests/definitions.go
```diff
@@ -30,5 +30,10 @@ type CrdFilter struct {
...
+	ItrSpecFilter string   //Filter is appended with "=="<some-crd-name>")]", so no need to write it completely. Just write the part before double equals
+	VField        string
```
leecalcote, line 35: Could you toss in a brief comment to describe these variables?

utils/manifests/definitions.go
Revolyssup: Iterator
It Iterates over the objects and picks only the one we are interested in

utils/manifests/definitions.go
Revolyssup: Itr filter is for defintions
ItrSpec filter is for Schemas
The itr filter has to be postfixed by "==)]" and is used when we are filtering for certain specific crds or resources.
So we iterate over the entire json and extract only the objects we are interested in.

leecalcote (approved):

utils/manifests/definitions.go
utkarsh-pro: @Revolyssup, I think @leecalcote meant to add some comments in the code. Right now the field names are too cryptic.  It would be come back to bite us when someone else would try to use the struct.

utils/manifests/definitions.go
```diff
@@ -30,5 +30,10 @@ type CrdFilter struct {
...
+	GField        string
+	IsJson        bool //Set to true if input format is json
```
utkarsh-pro, line 37: Consider document like this, it is picked up by the intellisense as well.
```suggestion
        // IsJson should be set to true if the input format is json. Defaults to false
	IsJson        bool
```

utils/manifests/definitions.go
Revolyssup: Okay will do

Revolyssup:
@utkarsh-pro  Have added verbose comments now

utkarsh-pro (approved):

PR #110, opened 2021-09-29
Changed kubeopenapi-jsonschema version to latest.

Revolyssup:
**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/utils/manifests/generateComponent.go
+++ b/utils/manifests/generateComponent.go
@@ -20,7 +20,7 @@ func GenerateComponents(manifest string, resource int, cfg Config) (*Component, 
 	wd := filepath.Join(utils.GetHome(), ".meshery", "bin")
 	fmt.Println("Looking for kubeopenapi-jsonschema in ", wd)
 	var binPath string = filepath.Join(wd, "kubeopenapi-jsonschema")
-	var url string = "https://github.com/layer5io/kubeopenapi-jsonschema/releases/download/v0.1.0/kubeopenapi-jsonschema"
+	var url string = "https://github.com/layer5io/kubeopenapi-jsonschema/releases/download/v0.1.2/kubeopenapi-jsonschema"
 	switch runtime.GOOS {
 	case "windows":
 		binPath += ".exe"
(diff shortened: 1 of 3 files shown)


utils/manifests/generateComponent.go
```diff
@@ -20,7 +20,7 @@ func GenerateComponents(manifest string, resource int, cfg Config) (*Component,
 	wd := filepath.Join(utils.GetHome(), ".meshery", "bin")
 	fmt.Println("Looking for kubeopenapi-jsonschema in ", wd)
 	var binPath string = filepath.Join(wd, "kubeopenapi-jsonschema")
-	var url string = "https://github.com/layer5io/kubeopenapi-jsonschema/releases/download/v0.1.0/kubeopenapi-jsonschema"
+	var url string = "https://github.com/layer5io/kubeopenapi-jsonschema/releases/download/v0.1.2/kubeopenapi-jsonschema"
```
DelusionalOptimist, line 23: we might want to fetch the latest release programmatically and not hard code it :sweat_smile:

DelusionalOptimist: there is code for this in adapters which could be made quick work of, if we're willing to do so in this PR itself: https://github.com/meshery/meshery-istio/blob/master/internal/config/releases.go

Revolyssup: I have done this knowingly, to avoid breaking changes in meshery in case something incompatible takes place in the latest version.

DelusionalOptimist: hmm... sure :+1:

leecalcote: The issue is that we're not tracking versions here. This is off our radar. As the primary and really only intended consumers of this tool, we should (and will) only make releases when they are intended for this project's consumption.

Let's think aloud about possible approaches to both avoid bugs based on releases with breaking changes and not knowing that we've hardcoded a version here (out of the 60+ repos that we have with the thousands of dependencies that they have).

leecalcote: Issue #110 created to follow up on this.

leecalcote (approved):
merging to allow k8s component progress to continue.

leecalcote:
Uh-oh. I'm just seeing the do-not-merge label. What else other than @DelusionalOptimist's comment is the merge of this issue dependent upon?

PR #125, opened 2021-11-03
[Chore] Shifting meshery functions to meshkit

piyushsingariya:
Signed-off-by: Piyush Singariya <piyushsingariya@gmail.com>

**Description**

This PR introduces changes shifting a few functions from mesheryctl to meshkit.

This PR fixes #

**Notes for Reviewers**


utils/kubernetes/healthcheck.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	if (minimalRequirementVersion[0] == actualVersion[0]) && (minimalRequirementVersion[1] == actualVersion[1]) && (minimalRequirementVersion[2] <= actualVersion[2]) {
+		return true
+	}
+
+	return false
+}
+
+var semVer = regexp.MustCompile("(v[0-9]+.[0-9]+.[0-9]+)")
```
hexxdump, line 99: @piyushsingariya, Any reason for having this outside the parseKubectlShortVersion() ? Is there a plan to use it in any other function? If not, we can move it inside the function.

piyushsingariya: yeah sure, good idea.

utkarsh-pro:
@piyushsingariya, I might be missing the context here but why are we moving these functions to meshkit?

utkarsh-pro:
CheckKubectlVersion appears to be a function that is very specific to mesheryctl as no other meshery component expects to have access to that binary, won't we be better off having it in mesheryctl only then?

piyushsingariya:
> @piyushsingariya, I might be missing the context here but why are we moving these functions to meshkit?

these functions are interacting with other software packages (k8s, and kubectl here) and hence making it hard to test these during the workflows. long time ago I was told to shift the complete file to meshkit(and It will help in codecov) so I thought to raise the PR.



> CheckKubectlVersion appears to be a function that is very specific to mesheryctl as no other meshery component expects to have access to that binary, won't we be better off having it in mesheryctl only then?

True, this function is specific, and can be kept to mesheryctl, if not marked as an exception.

utkarsh-pro:
@piyushsingariya I am not quite sure about this PR. We usually put the functions in meshkit which are supposed to be shared by meshery server, mesheryctl, adapters, meshsync, etc. While we put the functions that will be shared among adapters in meshery-adapter-library. 

I don't see these functions falling in either of those two categories.

piyushsingariya:
> @piyushsingariya I am not quite sure about this PR. We usually put the functions in meshkit which are supposed to be shared by meshery server, mesheryctl, adapters, meshsync, etc. While we put the functions that will be shared among adapters in meshery-adapter-library.
> 
> I don't see these functions falling in either of those two categories.

Let's close if someone can +1 to this.

hexxdump:
> @piyushsingariya I am not quite sure about this PR. We usually put the functions in meshkit which are supposed to be shared by meshery server, mesheryctl, adapters, meshsync, etc. While we put the functions that will be shared among adapters in meshery-adapter-library.
> 
> I don't see these functions falling in either of those two categories.

I agree some of the the functions are very specific to mesheryctl. However some (e.g. GetK8sVersionInfo(), getK8sVersion () ) can be leveraged by any meshery component.

hexxdump (approved):
I am fine with the code refactoring.

leecalcote:
What @utkarsh-pro said is accurate. 

Meshery Server does need to interface with Kubernetes clusters a fair bit. Currently, we don't have any cluster checks beyond the ability perform an ad hoc Kubernetes API connectivity test. 

Is there an alternate location that we can lodge this in mesheryctl until such time as Meshery Server finds itself in need of these same checks?

No comment, @navendu-pottekkat?

piyushsingariya:
Closing this, I also agree, that there isn't much point in moving them to meshkit.

PR #133, opened 2021-12-06
Add utility function for sorting un-normalized semantic versioning

Revolyssup:

**Description**

Across adapters as well as meshery server, a lot of times there comes the use case of sorting semantic versioning where we are not sure if the versions are normalized. This utility function does that
**Notes for Reviewers**

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,6 @@ require (
 	github.com/nats-io/nats.go v1.9.1
 	github.com/onsi/ginkgo v1.14.1 // indirect
 	github.com/onsi/gomega v1.10.2 // indirect
-	github.com/pkg/errors v0.9.1
 	github.com/sirupsen/logrus v1.8.1
 	github.com/spf13/cobra v1.1.3
 	github.com/spf13/viper v1.8.1
--- a/utils/sortversions.go
+++ b/utils/sortversions.go
@@ -1,0 +1,61 @@
+package utils
+
+import (
+	"sort"
+	"strconv"
+	"strings"
+)
+
+type dottedStrings []string
+
+func (d dottedStrings) Less(i, j int) bool {
+	si := d[i]
+	sj := d[j]
+	si = cleanup(si)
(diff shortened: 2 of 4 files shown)


leecalcote:
Any need for error handling here? Handling of invalid arguments non-semantic version-matching strings?

Revolyssup:
> Any need for error handling here? Handling of invalid arguments non-semantic version-matching strings?

This function internally normalizes into semantic version form even if the passed string was bad. I dont think we need an error here because if the passed strings are completely non-sensicle, it will still sort it as close as possible. Any malformed characters will be ignored internally

leecalcote:
I wonder if there isn't an existing library for this.

Can you expand on the need for this beyond "we're not sure if the versions are normalized". Is this for purposes of identifying latest version and ensuring that we're comparing correctly by ensuring that the versions are correctly sorted (in order)?

Revolyssup:
> I wonder if there isn't an existing library for this.
> 
> Can you expand on the need for this beyond "we're not sure if the versions are normalized". Is this for purposes of identifying latest version and ensuring that we're comparing correctly by ensuring that the versions are correctly sorted (in order)?

The recently added getlatestrelease function works just fine, except for this-
Sometime a repo makes new release and after some time, falls back to an older release as latest/stable for some reason. For eg- Istio had 1.12.0 as latest but now it has fallen back to 1.11.5.
So what should we consider the latest? 1.12.0 or 1.11.5? I think 1.12.0
But this function doesn't help us to get that. By scraping we can get first 5-10 latest versions which could be in any form, normalized or not, and we would need to find the largest(latest) of them. So I wrote this for that particular use case

Revolyssup:
@leecalcote  Also I did find already existing semver package for that. But it would work strictly with versions like 1.0.2,3.0.4. Even "v1.4.1" makes it panic. It is very unreliable and can cause random panics. My function on the other hand will never panic.

utkarsh-pro (approved):
LGTM

PR #141, opened 2022-01-02
Add DBClose method

Revolyssup:

**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/database/database.go
+++ b/database/database.go
@@ -32,6 +32,17 @@ type Handler struct {
 	// Implement methods if necessary
 }
 
+func (h *Handler) DBClose() error {
+	db, err := h.DB.DB()
+	if err != nil {
+		return err
+	}
+	err = db.Close() //It ensures that all writes have completed and the database is not corrupted.
+	if err != nil {
+		return err
+	}
+	return nil
+}
 func New(opts Options) (Handler, error) {
 	switch opts.Engine {
 	case POSTGRES:
@@ -46,6 +57,7 @@ func New(opts Options) (Handler, error) {
 		if err != nil {
 			return Handler{}, ErrDatabaseOpen(err)
 		}
+
 		return Handler{
 			db,
 			&sync.Mutex{},
--- a/database/error.go
+++ b/database/error.go
@@ -3,16 +3,16 @@ package database
 import "github.com/layer5io/meshkit/errors"
 
 var (
-	ErrNoneDatabaseCode           = "11010"
-	ErrDatabaseOpenCode           = "11011"
-	ErrSQLMapUnmarshalJSONCode    = "11012"
-	ErrSQLMapUnmarshalTextCode    = "11013"
-	ErrSQLMapMarshalValueCode     = "11014"
-	ErrSQLMapUnmarshalScannedCode = "11015"
-	ErrSQLMapInvalidScanCode      = "11016"
-
-	ErrNoneDatabase      = errors.New(ErrNoneDatabaseCode, errors.Alert, []string{"No Database selected"}, []string{}, []string{"database name is empty"}, []string{"Input a name for the database"})
-	ErrSQLMapInvalidScan = errors.New(ErrSQLMapUnmarshalScannedCode, errors.Alert, []string{"invalid data type: expected []byte"}, []string{}, []string{}, []string{})
+	ErrNoneDatabaseCode              = "11010"
+	ErrDatabaseOpenCode              = "11011"
+	ErrSQLMapUnmarshalJSONCode       = "11012"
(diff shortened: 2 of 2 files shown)


database/error.go
```diff
@@ -3,16 +3,16 @@ package database
...
+	ErrNoneDatabaseCode              = "11010"
+	ErrDatabaseOpenCode              = "11011"
+	ErrSQLMapUnmarshalJSONCode       = "11012"
+	ErrSQLMapUnmarshalTextCode       = "11013"
+	ErrSQLMapMarshalValueCode        = "11014"
+	ErrSQLMapUnmarshalScannedCode    = "11015"
+	ErrSQLMapInvalidScanCode         = "11016"
+	ErrClosingDatabaseConnectionCode = "11017"
+	ErrNoneDatabase                  = errors.New(ErrNoneDatabaseCode, errors.Alert, []string{"No Database selected"}, []string{}, []string{"database name is empty"}, []string{"Input a name for the database"})
+	ErrSQLMapInvalidScan             = errors.New(ErrSQLMapUnmarshalScannedCode, errors.Alert, []string{"invalid data type: expected []byte"}, []string{}, []string{}, []string{})
```
leecalcote, line 15: Please enhance this error message.

database/error.go
```diff
@@ -42,3 +42,8 @@ func ErrSQLMapMarshalValue(err error) error {
 func ErrSQLMapUnmarshalScanned(err error) error {
 	return errors.New(ErrSQLMapUnmarshalScannedCode, errors.Alert, []string{"failed to unmarshal scanned data", err.Error()}, []string{err.Error()}, []string{}, []string{})
 }
+
+//ErrClosingDatabaseConnection represents the error which will occur when the database connection fails to get closed
+func ErrClosingDatabaseConnection(err error) error {
+	return errors.New(ErrClosingDatabaseConnectionCode, errors.Alert, []string{"failed to close database connection"}, []string{err.Error()}, []string{"Invalid database instance passed."}, []string{"Make sure the DB handler has a valid database instance."})
```
leecalcote, line 48: Is this guaranteed to complete when a kill signal is received? If not, the message might need to consider for this.

Also, if not, we need to consider for that in our database connections.

database/error.go
Revolyssup: You mean rename the error code? What is there to enhance exactly here? I didn't quite get you

database/error.go
Revolyssup: In the dev environment, yes it will definitely complete after ctrl+C.  I think this will return easily everytime the os.Interrupt is called.

database/error.go
leecalcote: Oh, a probable cause and suggested remediation are missing.

database/error.go
leecalcote: Sounds good.

Revolyssup: Although if the shell is closed directly, which is the parent process of the running go process, that cannot be handled.

database/error.go
Revolyssup: My PR only added line 13, which is just a code. And the error that I added does have both of the asked things.

leecalcote: Oh, good. Ok. You're directly off the hook here, then.

Indirectly, you're on the hook. I encourage you to think holistically for the health of the project.

database/error.go
leecalcote: "The shell" being a contributor's shell session or being the Docker container's base process?

Revolyssup: Yes.

database/database.go
```diff
@@ -32,6 +32,17 @@ type Handler struct {
 	// Implement methods if necessary
 }
 
+func (h *Handler) DBClose() error {
```
leecalcote, line 35: Is it reasonable that this function be invoked when a SIGTERM is trapped?

Revolyssup: I think yes, it is reasonable. SInce the default time betweent first SIGTERM and SIGKILL is 30 seconds in kubernetes, that is more than enough to close the connections if the pod terminates. Same with docker stop.

leecalcote: And for Docker, it's 10 seconds, I believe.

Good. Please make it so.

Revolyssup: After this is merged, when I will create the PR in Meshery I will add SIGTERM as one of the signals on which the process will trap.

leecalcote: Perfect. Thank you, @Revolyssup

leecalcote (approved):

database/database.go
leecalcote: And add this function that trap?

PR #152, opened 2022-02-01
adding output option for logger for golang testing in mesheryctl

piyushsingariya:
Signed-off-by: Piyush Singariya <piyushsingariya@gmail.com>

**Description**

This PR adds a new option for adding a custom output writer which will help in golang tests for mesheryctl where logs are captured via a string buffer.

**Notes for Reviewers**

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -54,8 +54,11 @@ func New(appname string, opts Options) (Handler, error) {
 
 	// log.SetReportCaller(true)
 	log.SetOutput(os.Stdout)
-	log.SetLevel(logrus.InfoLevel)
+	if opts.Output != nil {
+		log.SetOutput(opts.Output)
+	}
 
+	log.SetLevel(logrus.InfoLevel)
 	if opts.DebugLevel {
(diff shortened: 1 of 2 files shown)


logger/logger.go
```diff
@@ -54,8 +54,11 @@ func New(appname string, opts Options) (Handler, error) {
...
 	// log.SetReportCaller(true)
 	log.SetOutput(os.Stdout)
-	log.SetLevel(logrus.InfoLevel)
+	if opts.Output != nil {
+		log.SetOutput(opts.Output)
+	}
 
+	log.SetLevel(logrus.InfoLevel)
 	if opts.DebugLevel {
 		log.SetLevel(logrus.DebugLevel)
```
leecalcote, line 63: Do you have an example of how / where Debug will be flagged as "on"/"true"?

piyushsingariya: In mesheryctl we need Debug to be turned on, as many times we print additional info with the flag `--verbose`

leecalcote (approved):

PR #174, opened 2022-04-19
fixed Helm issue where namespace was not persisted

jbyers19:
Signed-off-by: Jared Byers <j.byers@f5.com>

**Description**
Updated the way the Helm action config was being created when applying Helm charts.

This PR fixes https://github.com/meshery/meshery-nginx-sm/issues/39

**Notes for Reviewers**
This change is to fix the issue where a mesh is deployed in the `meshery` namespace regardless of which namespace is specified when initiating the deploy sequence. The root cause is Helm always using the namespace provided in the initial config, which must be `meshery` because that's where mesh resources were being deployed.

Tested this change with Istio, Linkerd, and NGINX adapters and it worked as expected.

- [X] Yes, I signed my commits.

### DIFF
--- a/utils/kubernetes/apply-helm-chart.go
+++ b/utils/kubernetes/apply-helm-chart.go
@@ -12,9 +12,9 @@ import (
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/chart/loader"
+	"helm.sh/helm/v3/pkg/cli"
 	"helm.sh/helm/v3/pkg/getter"
 	"helm.sh/helm/v3/pkg/repo"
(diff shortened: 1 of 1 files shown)


piyushsingariya:
@jbyers19 can you also make sure that meshery installation also work as expected with mesheryctl with this update??

jbyers19:
> @jbyers19 can you also make sure that meshery installation also work as expected with mesheryctl with this update??

Yes. I tested it and it worked.

piyushsingariya (approved):
CITY! 😎

alphaX86 (approved):
Looks good on my side 👍

leecalcote:
> Tested this change with Istio, Linkerd, and NGINX adapters and it worked as expected.

Awesome. Thanks for running through a few of these to verify, @jbyers19

leecalcote (approved):

PR #188, opened 2022-06-29
Adds support for dictionary to enhance formatting

theBeginner86:
**Description**

Screenshot 2022-06-29 at 7 29 20 PM

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/utils/manifests/utils.go
+++ b/utils/manifests/utils.go
@@ -197,17 +197,6 @@ func deleteFile(path string) error {
 func useDictionary(input string) string {
 	dic := map[interface{}]interface{} {  // includes Whitelist words
 		"Mesh Sync": "MeshSync",
-		"api"      : "API",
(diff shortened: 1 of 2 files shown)


utils/manifests/utils.go
```diff
@@ -193,6 +193,19 @@ func deleteFile(path string) error {
 	return nil
 }
 
+// This helps in formating the leftover fields using a pre-defined dictionary
+func useDictionary(input string) string {
+	dict := map[interface{}]interface{}{ // includes Whitelist words
```
sudo-NithishKarthik, line 198: why not just map[string]string ?

theBeginner86: I thought with that we can supply more general types. But using map[string]string would also be fine as we would just be using strings types.
I'll make that change.

PR #197, opened 2022-07-28
Refactor controller status logic

MUzairS15:
Signed-off-by: MUzairS15 <muzair.shaikh810@gmail.com>

**Description**

This PR refactor the logic of status of Meshery Components wrt this [doc](https://docs.google.com/document/d/1q-aayRqx3QKIk2soTaTTTH-jmHcVXHwNYFsYkFawaME/edit#heading=h.ngupcd4j1pfm).
**Notes for Reviewers**


MUzairS15:
`GetPublicEndpoint` doesn't seem to work correctly, as even if broker-service has IP as pending this function returns an endpoint.
@leecalcote @sudo-NithishKarthik Can you explain this behaviour??

leecalcote:
@MUzairS15 what is the output of the function in that scenario?

leecalcote:
@sudo-NithishKarthik please review.

MUzairS15:
> @MUzairS15 what is the output of the function in that scenario?
Screenshot 2022-07-28 at 8 35 07 PM
The output is the value in `External` field, though svc is not exposed.

MUzairS15:
I think er should query for broker-svc instead of custom resource

leecalcote:
> I think we should query for broker-svc instead of custom resource

Yes, potentially both. You're right that the service is ultimately what we need working / connected.

leecalcote:
Starting with the broker svc is the right thing, though.

models/controllers/meshery_broker.go
```diff
@@ -100,5 +104,22 @@ func (mb *mesheryBroker) GetPublicEndpoint() (string, error) {
...
+		return getImageVersionOfContainer(statefulSet.Spec.Template, "nats"), nil
+	}
+	return mb.version, nil
+}
+
+func getImageVersionOfContainer(container v1.PodTemplateSpec, containerName string) string {
+	var version string
+	for _, container := range container.Spec.Containers {
+		if strings.Compare(container.Name, containerName) == 0 {
+			version = strings.Split(container.Image, ":")[1]
```
Revolyssup, line 121: Although there will be a tag, but still avoid accessing an index before checking length of array

MUzairS15:
@Revolyssup

sudo-NithishKarthik:
> GetPublicEndpoint doesn't seem to work correctly, as even if broker-service has IP as pending this function returns an endpoint.
@leecalcote @sudo-NithishKarthik Can you explain this behavior??


This might be because we have fallbacks that give the internal endpoint when it could not find the external endpoint

MUzairS15:
> > GetPublicEndpoint doesn't seem to work correctly, as even if broker-service has IP as pending this function returns an endpoint.
> > @leecalcote @sudo-NithishKarthik Can you explain this behavior??
> 
> This might be because we have fallbacks that give the internal endpoint when it could not find the external endpoint

+1

sudo-NithishKarthik:
In that case, we can either modify the function to not give internal endpoint or to return that information

models/controllers/helpers.go
```diff
@@ -13,6 +16,16 @@ import (
 	"github.com/spf13/viper"
 )
 
+const BrokerPingEndpoint = "8222/connz"
+
+type Connections struct {
```
Revolyssup, line 21: Where will this struct be used?

Revolyssup:
@MUzairS15 Also the lint tests failing

models/controllers/helpers.go
MUzairS15: https://github.com/meshery/meshkit/pull/197/files#diff-a127b332cfc1dd1adf96e28240276fe52634ce187468c511bf31d30d55321a0eR112

Revolyssup (approved):
Just fix the lint checks and we are good to go

PR #206, opened 2022-08-16
[Feat] Component Generation Factory

sudo-NithishKarthik:
**Description**

This PR creates a Component Generation factory which we can use to create generator functions that are capable of generating Meshery Components from various sources. This PR also includes some refactoring on the old component generation codebase. 

 Spec: [Component Generation Factory](https://docs.google.com/document/d/1B2N78EdRiZF-yVo1-HY3syppwBBDumgMuYg6seD-AJ4/edit#heading=h.g23s20igxoxk)


leecalcote:
Failing lint checks (outside of this PR, I think, but still need to be addressed) - https://github.com/meshery/meshkit/runs/7858021156?check_suite_focus=true#step:4:79

leecalcote (changes requested):
@sudo-NithishKarthik, please address all lint item inside of and outside of this PR.

Revolyssup:
@sudo-NithishKarthik  The methods that are implemented by ArtifactHUb struct can be turned into a generic interface and implemented by artifact hub struct. Currenly we only scan artifact hub, support for other registries could be added later with uniformity. Also the Generate function, just takes the `name` right now. It would be Nice to have that encapsulated in a struct so that it is easier to extend the functionality without breaking the API contract and have some backwards compatibility.
Suggestion for that struct: Have name as a field and also metadata. Some use case can be, scanning through protected registries might require some token, which then can be encapsulated inside metadata.

PR #221, opened 2022-10-06
Fix/Hack for helm cli bug[In progress][Do not merge]

Revolyssup:
Signed-off-by: revolyssup <ashishjaitiwari15112000@gmail.com>

**Description**
1. Helm cli doesn't respect the passed kubeconfig and always defaults to the one in ~/.kube/config and in the case of in-cluster deployments it breaks as there is nothing inside ~/.kube/config
This PR sets the appropriate env variable through which helm cli can consume our input rather than setting it to empty in all cases. 
2. Also the applyOperatorHelmChart function wasn't undeploying operator for in-cluster deployments but rather tried to upgrade the operator. This weird logic might have crept in some time ago so this PR also fixed that.
**Notes for Reviewers**

### DIFF
--- a/models/controllers/helpers.go
+++ b/models/controllers/helpers.go
@@ -13,7 +13,6 @@ import (
 	"github.com/layer5io/meshery-operator/api/v1alpha1"
 	"github.com/layer5io/meshkit/utils"
 	mesherykube "github.com/layer5io/meshkit/utils/kubernetes"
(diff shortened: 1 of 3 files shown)


Revolyssup:
This PR doesn't gaurantee a complete fix for now but still fixes a bug so merging this. I will update on the progress here.

PR #225, opened 2022-10-23
[Refactor] Update ArtifactHub component generation logic 

sudo-NithishKarthik:
**Description**

This PR 

- Updates the logic for generating components from ArtifactHub to make use of the [helmExporterDump](https://artifacthub.io/docs/api/#/Integrations/getHelmExporterDump) integration
- This makes the generation process much faster and helps us escape the rate-limiting issue since we now just have to send 1 request in place of 150
- It makes use of the fact that we can access the list of charts of a repository via appending `/index.yaml` on `repository_url` which is given as a package metadata
- Adds a function that generates AhPackages of all HELM charts available in ArtifactHub
- Adds support for use of multiple configs when extracting schema from CRDs. It is needed since the format in which the CRDs are structured slightly varies from system to system and we need the ability to accommodate that. 

**Notes for Reviewers**

### DIFF
--- a/cmd/syncmodutil/internal/modsync/sync.go
+++ b/cmd/syncmodutil/internal/modsync/sync.go
@@ -37,14 +37,14 @@ type GoMod struct {
 	RequiredVersions []Package
 }
 
-//For debugging
+// For debugging
 func (g *GoMod) PrintRequiredVersions() {
 	for _, v := range g.RequiredVersions {
 		fmt.Printf("package=%s ;version=%s\n", v.Name, v.Version)
 	}
 }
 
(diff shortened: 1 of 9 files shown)


leecalcote (changes requested):
A few lint check items to review...

leecalcote:
I ❤️ this write up in the issue description.

leecalcote:
Does the failed check with missing test files for ```github.com/layer5io/meshkit/utils/walker``` need to be separately addressed?

sudo-NithishKarthik:
Yes

PR #227, opened 2022-10-28
Add Database compliant Meshmodel structs

Revolyssup:
Signed-off-by: revolyssup <ashishjaitiwari15112000@gmail.com>

**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/models/meshmodel/core/v1alpha1/component.go
+++ b/models/meshmodel/core/v1alpha1/component.go
@@ -9,20 +9,30 @@ type TypeMeta struct {
 
 // use NewComponent function for instantiating
 type Component struct {
-	TypeMeta
-	ComponentSpec
-	Metadata map[string]interface{} `json:"metadata,omitempty"`
+	TypeMeta      `gorm:"embedded"`
+	ComponentSpec `gorm:"embedded"`
+	Metadata      map[string]interface{} `json:"metadata,omitempty" gorm:"type:JSONB"`
 	// for backward compatibility
 	Spec string `json:"spec,omitempty"`
 }
-
+type capability struct {
+	ID string `json:"id,omitempty"`
+	// Host is the address of the service registering the capability
+	Host string `json:"host,omitempty"`
+}
 type ComponentSpec struct {
-	Schematic map[string]interface{} `json:"schematic,omitempty"`
+	Schematic map[string]interface{} `json:"schematic,omitempty" gorm:"type:JSONB"`
 }
 
 func NewComponent() Component {
 	comp := Component{}
 	comp.APIVersion = "core.meshery.io/v1alpha1"
 	comp.Kind = ComponentDefinitionKindKey
+	comp.Metadata = make(map[string]interface{}, 1)
 	return comp
+}
+
+type ComponentCapability struct {
+	Component
+	capability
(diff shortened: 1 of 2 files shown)


models/meshmodel/core/v1alpha1/component.go
```diff
@@ -9,20 +9,30 @@ type TypeMeta struct {
 
 // use NewComponent function for instantiating
 type Component struct {
-	TypeMeta
-	ComponentSpec
-	Metadata map[string]interface{} `json:"metadata,omitempty"`
+	TypeMeta      `gorm:"embedded"`
+	ComponentSpec `gorm:"embedded"`
+	Metadata      map[string]interface{} `json:"metadata,omitempty" gorm:"type:JSONB"`
```
leecalcote, line 14: 😍 JSONB

models/meshmodel/core/v1alpha1/component.go
```diff
@@ -9,20 +9,30 @@ type TypeMeta struct {
...
 	// for backward compatibility
 	Spec string `json:"spec,omitempty"`
 }
-
+type capability struct {
```
leecalcote, line 18: Please verify that these structs match the Meshery object model table designs.

leecalcote (changes requested):

models/meshmodel/core/v1alpha1/component_capabilities_db.go
```diff
@@ -0,0 +1,51 @@
...
+// This file consists of methods and structs that database(gorm) will use to interact with meshmodel components
+type ComponentDB struct {
+	TypeMeta
+	ComponentSpecDB
+	Metadata []byte `json:"metadata"`
+	// for backward compatibility
+	Spec string `json:"spec,omitempty"`
+}
+
+type ComponentSpecDB struct {
```
leecalcote, line 18: Please name the structs with the same table names from the object model.

Revolyssup: Actually there will be two structs. One used to serve client response `ComponentSpec` and another used to create table in the database `ComponentSpecDB`. The reason is that map[string]interface{} is incompatible with sql database. And even if I make the `metadata` field - []byte, the client will not see readable response. This similar approach is already present in some SMP related struct in meshery.

models/meshmodel/core/v1alpha1/component.go
Revolyssup: This struct was present in meshery/meshery and is being moved here for a central source of truth.

models/meshmodel/core/v1alpha1/component_capabilities_db.go
Revolyssup: Actually this one is a helper struct. `DB` sufficed structs are helper structs which represent the data in the database. Because map[string]interface{} can not be pushed inside sql database and converting it into any other type such as []byte will produce an output unreadable by the client. So the client facing struct is different than this one, This approach is already used in meshery at one other place. It's a little awkward but there is no other way, I could find in my research.

sudo-NithishKarthik (approved):
LGTM

PR #247, opened 2023-01-03
[Refactor] Add `ErrorV2` struct 

sudo-NithishKarthik:
**Description**

 There are some limitations with the existing `Error` type:
1. There are different types of Errors. Each type of error contains different information which cannot be contained in the `Error` struct.
2. Short and Long Descriptions accept only strings and so they cannot contain structured information.
3. For example, the Validation Error information i.e. `instancePath`, `badValue`, etc. cannot be contained in the `Error` struct (or not in a structured way that the clients can make use of it).

`ErrorV2` Proposal: 
1. `ErrorV2` struct adds the ability to express all different types of errors.
2. It does so via adding a new field to the struct called `AdditionalInfo` which can contain structured error-specific information.
3. `ErrorV2` is backward compatible with the `Error` struct.

**Notes for Reviewers**

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -83,6 +83,18 @@ func New(code string, severity Severity, sdescription []string, ldescription []s
 	}
 }
 
+func NewV2(code string, severity Severity, sdescription []string, ldescription []string, probablecause []string, remedy []string, additionalInfo interface{}) *ErrorV2 {
+	return &ErrorV2{
+		Code:                 code,
+		Severity:             severity,
+		ShortDescription:     sdescription,
+		LongDescription:      ldescription,
+		ProbableCause:        probablecause,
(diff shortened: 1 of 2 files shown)


leecalcote:
How does this interplay with Meshery's use of CloudEvents?

leecalcote:
What facility, if any, does this offer to chain a set of errors together?

leecalcote:
We're in need of a "type" and/or "category" attribute.

sudo-NithishKarthik:
Let's use `CloudEvents`

sudo-NithishKarthik:
Reopening this since even with `CloudEvents`, we will have to define such a structure ourselves since `CloudEvents` does not define such domain-specific things.

PR #268, opened 2023-02-20
do not register components with empty schemas

Revolyssup:
**Description**

This PR fixes #

**Notes for Reviewers**


leecalcote (commented):
@Revolyssup How many will we lose?

Revolyssup:
> @Revolyssup How many will we lose?

Bad news: In total 205 components ranging around 55 models.
Good news: For each of the 55 models, it isn't the case that all components have missing schema. Only few do. So all these 55 models are a mix of schema-less and schema-having components.

PR #285, opened 2023-04-01
compose helpers ADD

Azanul:
**Description**

This PR adds helper functions for using compose in Meshery & its components

**Notes for Reviewers**

- [X] Yes, I signed my commits.


Azanul:
@Revolyssup  @leecalcote

Azanul:
@hexxdump @acald-creator @alphaX86

helpers/compose.go
```diff
@@ -0,0 +1,166 @@
...
+			ID:      c.ID,
+			Image:   c.Image,
+			Status:  c.Status,
+			Command: c.Command,
+			Ports:   formatter.PortsToStrings(c.Ports, getFQDN(c)),
+		}
+	}
+	return retList
+}
+func getFQDN(container containers.Container) string {
```
hexxdump, line 120: Is there a specific need for this function?
How did we decide on which docker compose command to support/not support?
Is there some prior discussion, pls point me to it.

Azanul: > Is there a specific need for this function?

There isn't a specific need per se, I've tried to keep the implementation parallel to the [compose-cli](https://github.com/docker/compose-cli/blob/main/cli/cmd/ps.go)

> How did we decide on which docker compose command to support/not support?

I've analysed the usage in meshery to see which commands are needed, there's a PR there in which these functions will be used, replacing the `exec`s.

> Is there some prior discussion, pls point me to it.

There isn't. If you know of any compose commands used by any meshery component do let me know and I'll add those too.

helpers/compose.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	project := types.Project{Name: projectName, WorkingDir: filepath.Dir(composeFilePath), ComposeFiles: []string{composeFilePath}}
+
+	return c.ComposeService().Stop(ctx, &project, api.StopOptions{})
+}
+
+func Ps(ctx context.Context, c *client.Client, all, quiet bool, formatOpt string) error {
+	containerList, err := c.ContainerService().List(ctx, all)
+	if err != nil {
+		return errors.Wrap(err, "fetch containers")
```
hexxdump, line 78: can we use fmt.Errorf to be in line with the other error cases in this file ?

helpers/compose.go
hexxdump: So far I have seen instances of the below docker-compose commands in Meshery, so this should be fine. We can add more if required.

- up -d
- logs -f
- ps
- stop
- rm -f
- pull 

This LGTM.

helpers/compose.go
Azanul: Done

hexxdump (approved):

Azanul:
@leecalcote @acald-creator  Check it out

Issue #287, opened 2023-04-03
Writing Unit tests For k8 in Utils package 


Philip-21:
I  was thinking if its possible to write unit tests for each of the  [Kubernetes](https://github.com/meshery/meshkit/tree/master/utils/kubernetes) package code files. 
If it is i would like to engage in this

leecalcote:
Thank you, @Philip-21 👍 

// @Azanul @acald-creator @Revolyssup

Philip-21:
Ok @leecalcote  I will start working on it

Azanul:
@Aisuko This isn't completed

Philip-21:
@leecalcote please can you re- assign me to this ,  I mistakenly removed my assignment role . Thanks

leecalcote:
Thank you, @Aisuko 🙌

saurabh100ni:
This issue has been open for some time with no recent activity, unassigning to open it up for new contributors to give it a go.

ayushrakesh:
@saurabh100ni @leecalcote Please assign this issue to me.

ayushrakesh:
@Aisuko @saurabh100ni @Philip-21 Please review my PR.

ghost:
I'd like to work on this. Can I get assigned this please?

AnkitRewar11:
I would like to work on this issue. Could you please assign it to me?

PR #289, opened 2023-04-07
created unit test for the k8  describe package #287

Philip-21:
Signed-off-by: Philip-21 <philipuzomaobiora@gmail.com>

### Description 
The PR fixes for #287 

I created a k8 unit test using for the describe package . 

The role of mockClient is to provide a mock implementation of the meshkitkube.Client interface, which allows the <code>Describe()</code>  to be tested in isolation from the Kubernetes API. Instead of interacting with the real Kubernetes API, the  function can use the <code>Describe()</code> mock implementation provided by mockClient to simulate interactions with the API and verify that the function behaves correctly under different conditions.
The MockDescriber is used in the implementation of the <code>Describe()</code> to create a mock output that is returned when the
function is called with specific options
it takes in a Client object and a DescriberOptions object, and returns a string and an error

The approach of using a K8 api can slow down the test suite and make it more difficult to maintain.
Using mock objects and dependencies allows us to isolate the specific code we want to test and provide controlled inputs and outputs, without the need to interact with the actual Kubernetes API. This makes the tests faster, more reliable, and easier to maintain.
I added some comments where necessary to help understand the code better .


Philip-21:
I am also working on the other packages, i am still going through the code base to understand it better ,so i could write  the unit test.

utils/kubernetes/describe/describe_test.go
```diff
@@ -0,0 +1,88 @@
...
+			DescribeFunc: func(client *meshkitkube.Client, options DescriberOptions) (string, error) {
+				return "Name: test-deployment\nNamespace: test-namespace\n", nil
+			},
+			ExpectedOutput: "Name: test-deployment\nNamespace: test-namespace\n",
+			ExpectedError:  nil,
+		},
+	}
+
+	//run test cases
+	for _, tc := range testCases {
```
Aisuko, line 70: Thanks for your contribution. I beleive we really need unit test for all the repos. All the code is good and clean but there is consideration about using for run all the test cases manually.
Please check code below, we prefer use this way.

https://github.com/meshery/meshery-operator/blob/1ace1d307c2278325b3c7b9c741f7dc7af912529/api/v1alpha1/meshsync_types_test.go#L62-L95

utils/kubernetes/describe/describe_test.go
```diff
@@ -0,0 +1,88 @@
+package describe
+
+import (
+	"testing"
+
+	meshkitkube "github.com/layer5io/meshkit/utils/kubernetes"
+)
+
+/*
+  The MockDescriber is used in the implementation of the Describe function to create a mock output that is returned when the
```
Aisuko, line 10: Nice code comment and this is what we need. So, may you help add a code comment to the `describe.go` file. I believe if you have written a test case for it you have known what exactly it does.

Aisuko (changes requested):
There are two suggestions please take a look. And If we can make this happen in this repo. We can add test cases to other repos to increase the coverage.

Aisuko:
By the way, please sign off your commit, more detail https://github.com/meshery/meshkit/pull/289/checks?check_run_id=12627332331

utils/kubernetes/describe/describe_test.go
Philip-21: Sure, i can do this

utils/kubernetes/describe/describe_test.go
```diff
@@ -0,0 +1,88 @@
...
+			Options: DescriberOptions{
+				Name:      "test-pod",
+				Namespace: "test-namespace",
+				Type:      Pod,
+			},
+			/*
+				DescribeFunc field is a function that takes in a Client
+				object and a DescriberOptions object, and returns a string and an error
+			*/
+			DescribeFunc: func(client *meshkitkube.Client, options DescriberOptions) (string, error) {
```
Revolyssup, line 48: What is it that we are testing over here? You can't define the function that is to be tested inside of the test. This function is to be tested https://github.com/meshery/meshkit/blob/3f37557a2898d7de8a9548893302ca13e25358df/utils/kubernetes/describe/describe.go#L91
Ideally, you need to create a mock http server using httpmock and configure it to mock kube api server's response.

Revolyssup (changes requested):

utils/kubernetes/describe/describe_test.go
Philip-21: Yeah thats true i will remove the DescribeFunc field , to follow best test practice .
 
thanks for highlighting this i  will create mock http servers for k8 api response

Philip-21: The <code>DescribeFunc</code> wasnt used for testing it was the <code>tc.Options</code> i used as a parameter for testing. Instead <code>DescribeFunc()</code>  is a testcase field that creates a MockDescriber method for testing. 
The <code>func (m *MockDescriber) Describe(client *meshkitkube.Client, options DescriberOptions) (string, error) {
	return m.DescribeFunc(client, options)
}</code>
 simply returns the result of the <code>DescribeFunc()</code> instance

Philip-21: Ok thanks for the suggestion, i would create a mock http server response for the k8 api.

Aisuko: I agree with you here but I believe these codes are still useful. If we assume that `MockDescriber` has a function named `Describe` mock the real `Describe`returns it will be useful. Maybe he does not want to mock any of the API server's responses and only wants to test the function itself.  Although, it looks like the code does not test the target function features. We can change a little bit of this code to test the edge situation for the real `Describe` function.

Is that sound good guys?

Aisuko: Maybe we can discuss an outline for the mock test. It determines what's your target. After we define the scope, it would have a clear roadmap for mocking the test suite.

Philip-21: The approach of using a K8 api  to return any response can slow down the test suite and make it more difficult to maintain.
like @Revolyssup had said i should create a mock response for the api , which is what i will implement .

utils/kubernetes/describe/describe_test.go
Philip-21: ok this method seems new to me tho,  but i would see how i can implement it . I would reach out incase i have issues or i am stuck.

utils/kubernetes/describe/describe_test.go
Philip-21: Ok thanks I will look into it .

Aisuko: You are welcome. It is useful for all our repos.

leecalcote:
@Aisuko thanks for helping getting this PR into shape and ready for merge. Do you mind giving this another review?

@Philip-21 is there any outstanding feedback to address?

// @nebula-aac

Aisuko:
> @Aisuko thanks for helping getting this PR into shape and ready for merge. Do you mind giving this another review?
> 
> @Philip-21 is there any outstanding feedback to address?
> 
> // @nebula-aac

You are welcome @leecalcote. I saw that @Philip-21 remove all the files.

Philip-21:
> @Aisuko thanks for helping getting this PR into shape and ready for merge. Do you mind giving this another review?
> 
> @Philip-21 is there any outstanding feedback to address?
> 
> // @nebula-aac

 
This Pr had DCO and commit issues so I decided to remove all files
The issue is continued on #319 
I'm Still communicating with Ashish on the next step .

PR #311, opened 2023-05-09
[Policy] add rego policy support in meshkit

Abhishek-kumar09:
**Description**

This PR fixes #

1. Takes all the policy files from the directory and evaluates it against the given query.
2. The structs(in models) are the response structure that server should expect or marshalls/unmarshall into if required.

**Notes for Reviewers**


leecalcote:
@acald-creator no issues with dependencies here, right?

models/meshmodel/core/policies/data_models.go
leecalcote, line 1: @MUzairS15 @theBeginner86 will you help ensure that this struct is built from schema?

leecalcote (changes requested):

leecalcote:
Is this struct being defined in MeshKit so that the same struct can be used in Meshery Cloud or for another reason?

acald-creator:
reviewing

models/meshmodel/core/policies/data_models.go
leecalcote: @Abhishek-kumar09 is it?

Abhishek-kumar09: No. Its a response structure we would like to enforce. It contains the mix of information required for the javascript(frontend) and some about the port settings.

the `port` given it is `map[string]inerface{}` comes directly from what you provide in the design-file, thus following the schema rules.

Abhishek-kumar09: currently it is unused, the server would be the one to unmarshall the response from policy engine to this struct.

Abhishek-kumar09:
> Is this struct not being defined in Meshery Server, but in MeshKit so that the same struct can be used in Meshery Cloud or for another reason?

https://github.com/meshery/meshery/pull/7654#discussion_r1188501841
Since the policy engine is described here in the meshkit, its corresponding struct is also written here. But, it is meant to be used by the server, or any other client that could intake policy and expect some data-response from it.

models/meshmodel/core/policies/data_models.go
Abhishek-kumar09: Since now I get your questions after the development call, @MUzairS15 or @theBeginner86 if you guys have some design-spec or some research work already in place, lets get to it.

PR #325, opened 2023-07-10
Added count for unique models and components

abdullah1308:
**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/models/meshmodel/core/v1alpha1/component.go
+++ b/models/meshmodel/core/v1alpha1/component.go
@@ -90,7 +90,7 @@ func GetMeshModelComponents(db *database.Handler, f ComponentFilter) (c []Compon
 	countUniqueComponents := func(components []componentDefinitionWithModel) int {
 		set := make(map[string]struct{})
 		for _, model := range components {
-			key := model.ComponentDefinitionDB.Kind + "@" + model.ModelDB.Name
+			key := model.ComponentDefinitionDB.Kind + "@" + model.APIVersion + "@" + model.ModelDB.Name + "@" + model.ModelDB.Version
 			if _, ok := set[key]; !ok {
 				set[key] = struct{}{}
 			}
--- a/models/meshmodel/registry.go
+++ b/models/meshmodel/registry.go
@@ -194,8 +194,9 @@ func (rm *RegistryManager) GetModels(db *database.Handler, f types.Filter) ([]v1
 	countUniqueModels := func(models []modelWithCategories) int {
 		set := make(map[string]struct{})
 		for _, model := range models {
-			if _, ok := set[model.ModelDB.Name]; !ok {
-				set[model.ModelDB.Name] = struct{}{}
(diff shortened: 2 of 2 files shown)


leecalcote:
@abdullah1308 will you like into the Lint Checks?

models/meshmodel/core/v1alpha1/component.go
```diff
@@ -80,13 +80,24 @@ func CreateComponent(db *database.Handler, c ComponentDefinition) (uuid.UUID, er
...
 	type componentDefinitionWithModel struct {
 		ComponentDefinitionDB
 		ModelDB
 		CategoryDB
 	}
 
+	countUniqueComponents := func(components []componentDefinitionWithModel) int {
+		var set map[string]struct{}
+		for _, model := range components {
+			key := model.ComponentDefinitionDB.Kind + "@" + model.ModelDB.Name
```
leecalcote, line 93: Is this how duplicity is checked, @harkiratsm? With a composite of these three items, @MUzairS15?

zakisk: @abdullah1308 PR doesn't have link to issue where can I see issue?

zakisk:
@abdullah1308 PR title doesn't have a link to the issue where can I see the issue?

zakisk:
@abdullah1308 and can you please clarify to me, how the uniqueness of a component is defined. for example, if there is a slice of components like this `[A, C, B, A, D, F, E, C, P]` in your solution after loop iteration it ends up on the elements `[A, C, B, D, F, E, P]` contained in `set` map and returns  `len(set)` so is this desired uniqueness because the uniqueness can be evaluated as `[B, D, F, E, P]` since `A, C` are twice they aren't considered unique.

abdullah1308:
> @abdullah1308 and can you please clarify to me, how the uniqueness of a component is defined. for example, if there is a slice of components like this `[A, C, B, A, D, F, E, C, P]` in your solution after loop iteration it ends up on the elements `[A, C, B, D, F, E, P]` contained in `set` map and returns `len(set)` so is this desired uniqueness because the uniqueness can be evaluated as `[B, D, F, E, P]` since `A, C` are twice they aren't considered unique.

@zakisk Yes, that is how it works

models/meshmodel/core/v1alpha1/component.go
abdullah1308: @leecalcote [meshmap reference](https://github.com/layer5labs/meshery-extensions/blob/37bd8a7e82e27c133b1cb4cce238cadcac44cbdd/meshmap/src/utils/hooks/useMeshModelSearchModels.js#L43-L65)

abdullah1308: lint fixed

leecalcote (approved):

harkiratsm (commented):
Do we require changes in the meshery server? // @abdullah1308

PR #327, opened 2023-07-14
Register components based on ui compatibility

MUzairS15:
**Description**
Every component has `uiCompatible` property in their `metadata`. If `uiCompatible` property is unset then the component is still registered (if schema is valid) so that  `Applications` upload doesn't fail, but is not returned to UI.
`uiCompatible ` can be un-set in 2 cases:
1. Schema is not present,
2. UI cannot handle the schema correctly (eg: CRD).
**Notes for Reviewers**

### DIFF
--- a/models/meshmodel/core/v1alpha1/component.go
+++ b/models/meshmodel/core/v1alpha1/component.go
@@ -119,6 +119,9 @@ func GetMeshModelComponents(db *database.Handler, f ComponentFilter) (c []Compon
 			finder = finder.Where("component_definition_dbs.display_name = ?", f.DisplayName)
 		}
 	}
+	if !f.ReturnIncompatibleComponent {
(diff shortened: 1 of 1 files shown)


leecalcote (changes requested):
Where is the proposal for this new attribute `uiCompatible`?

MUzairS15:
> Where is the proposal for this new attribute `uiCompatible`?

Discussed here. https://app.clickup.com/t/85ztfxqh4?comment=90020034409921

leecalcote:
@MUzairS15 has this been separately addressed as `isAnnotation`?

leecalcote:
Close this PR?

MUzairS15:
No, this is different this was more about exposing control to user to selectively enable / disable meshmodels

PR #340, opened 2023-08-05
Set custom log level, and added formatting fields

nebula-aac:
**Description**

This PR adds
- formatted fields
- a custom log level to call

**Notes for Reviewers**


leecalcote:
Updating reviewers...

suhail34:
How can i test this changes in my local

Revolyssup:
@suhail34 You can checkout this branch locally and then in meshery/meshery, point your go.mod to your local meshkit using replace. Then you can start Meshery and see the logs. Since there are new functions that are added, you can try replacing current .Info, .Error functions with these new functions to see how the logs come out.


@nebula-aac It would be nicer if could have some unit tests here.

Revolyssup:
@nebula-aac Do you have some screenshot examples of how the error logs will look like running it locally?

suhail34:
I was thinking like if we can do the same as syslogformat and add in setformatter disableTimestamp: true something like this
```
TerminalLogFormat:
		log.SetFormatter(
&logrus.TextFormatter{
			DisableTimestamp: true
		}
)}
```

Revolyssup:
@suhail34 Also test it with Meshery server. I hope the logs don't get too big with 5-6 lines per error/warn log.

nebula-aac:
I'll have to make some updates on this PR. I'll provide some screenshots

nebula-aac:
If we want to keep it to the minimum on the logs for `mesheryctl`, I would like to know so that I'm not formatting the wrong fields.

MUzairS15:
@nebula-aac Any updates ?

suhail34:
I think we could define the init funciton and disable the timestamp so that globally all of the logrus has timestamp disabled

PR #35, opened 2021-03-01
service discovery modified

kumarabd:
Signed-off-by: kumarabd <abishekkumar92@gmail.com>

**Description**

Discovers service endpoint both internal and external.
First looks for load balancer in case of external service endpoint, and then nodeport.
**Notes for Reviewers**

**[Signed commits](../CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [x] Yes, I signed my commits.

PR #356, opened 2023-08-26
Add constructs for `Meshery Events`.

MUzairS15:
**Description**

This PR fixes #

**Notes for Reviewers**


MUzairS15:
https://github.com/meshery/meshkit/pull/35

models/events/build.go
```diff
@@ -0,0 +1,60 @@
...
+	operationId, _ := uuid.NewV4()
+	return &EventBuilder{
+		event: Event{
+			CreatedAt:   time.Now(),
+			OperationID: operationId,
+		},
+	}
+}
+
+func (e *EventBuilder) ActedOn(resource uuid.UUID) *EventBuilder {
```
leecalcote, line 23: ActedUpon, not "On"

models/events/build.go
```diff
@@ -0,0 +1,60 @@
...
+	e.event.Metadata = metadata
+	return e
+}
+
+func (e *EventBuilder) WithSeverity(severity EventSeverity) *EventBuilder {
+	e.event.Severity = severity
+	return e
+}
+
+func (e *EventBuilder) InitiatedBy(id uuid.UUID) *EventBuilder {
```
leecalcote, line 48: @MUzairS15 follow the table design spec exactly. Do not rename attributes.

leecalcote (changes requested):

models/events/build.go
MUzairS15: @leecalcote I have not diverged from the schema, the attribute name are exactly the same https://github.com/meshery/meshkit/pull/356/files/83dcd0952d500438cda2161e305b8ed1d12547a8#diff-5bc079c95a1811fd042d036db5a54fbae0ba3ef9e1645ce28d4eb4050b749049R49

models/events/build.go
MUzairS15: https://github.com/meshery/meshkit/pull/356/files/83dcd0952d500438cda2161e305b8ed1d12547a8#diff-5bc079c95a1811fd042d036db5a54fbae0ba3ef9e1645ce28d4eb4050b749049R24

models/events/build.go
leecalcote: Oh, I see. I'm confusing attribute name and function name. Symmetric function naming could be `fromUser` or InitiatingUser and InitiatingSystem.

models/events/build.go
leecalcote: Why not use the same moniker for both function and attribute like what is done with Description?

You're not being consistent with the approach. In some instances you are adding a verb or preposition to the function name, but not always. Consistency facilitates legibility and comprehension by your future self and others.

leecalcote (approved):
The only thing missing now is the split of `event_types`

PR #373, opened 2023-09-20
Implements handling of insecure TLS connections in Mehkit

yash37158:
**Description**

This PR fixes [#8619](https://github.com/meshery/meshery/issues/8619)

Implements handling of insecure TLS connections in Meshkit.

When connecting to Kubernetes clusters with a kubeconfig that has the insecure-skip-tls-verify: true setting configured, Meshery Server will now respect that setting and ignore TLS/SSL issues.

This feature is implemented in MeshKit and includes the following acceptance tests:

- Meshkit now enforces or ignores SSL errors on a per Kubernetes context basis, allowing import of a multi-context 
kubeconfig with the same or different insecure-skip-tls-verify settings per context.

- A manifest of all attempted, ignored, and successful connections are returned to the client in a Meshery Event (/api/events). But I still need a better understanding of how can i achieve this, on digging I came to know that Meshery uses pub/sub for passing events. but I am not sure if that is correct also how can I return it client(/api/event) and populate the ui? if someone would like to contribute or have a better approach for passing the status to the client.

Here is the test results for the desired state:
Screenshot 2023-09-21 at 12 51 12 AM

**Notes for Reviewers**


leecalcote:
Whoo-hoo, @yash37158 👍

utils/kubernetes/client.go
```diff
@@ -1,70 +1,162 @@
...
-	}
-
-	// Look for kubeconfig at the default path
-	path := filepath.Join(utils.GetHome(), ".kube", "config")
-	if config, err = clientcmd.BuildConfigFromFlags("", path); err == nil {
-		return config, err
-	}
-
-	return
+    if len(configfile) > 0 {
```
leecalcote, line 27: @yash37158 if you can eradicate the extraneous formatting changes, this will really help reviewers target considering your actual changes. Otherwise, review will take longer.

leecalcote (changes requested):

MUzairS15 (commented):
@yash37158 The current behaviour is if you provide the flag but your kube config has the 
`certificate-authority-data` for the cluster the `client-go` (k8s package) will ignore the flag, were as if you remove the `certificate-authority-data ` and then use the flag, it will be respected.

This behaviour is consistent with kubectl as well as k8s-client-go.

So I don't think this change is required Or you can do it in this way, if the flag is present remove the `certificate-authority-data`.

Philip-21:
> @yash37158 The current behaviour is if you provide the flag but your kube config has the `certificate-authority-data` for the cluster the `client-go` (k8s package) will ignore the flag, were as if you remove the `certificate-authority-data ` and then use the flag, it will be respected.
> 
> This behaviour is consistent with kubectl as well as k8s-client-go.
> 
> So I don't think this change is required Or you can do it in this way, if the flag is present remove the `certificate-authority-data`.

@MUzairS15 i don't think its best he removes the  `certificate-authority-data` . When users  are running mesheryctl on a k8 cluster, they may perform tasks that require  `certificate-authority-data`. . 
What do you think ??

Philip-21 (changes requested):
@yash37158 kindly work on your indents and newline spacing, meshkit is very sensitive to this.

MUzairS15:
@Philip-21 I agree, but the testcases he have, doesn’t include certificate-data hence the tests passes, (which comply with the default behaviour of client-go)
If we use the flag along with having certificate-data unless it is removed manually (which is not good) the behaviour will be overriden.

Philip-21:
> @Philip-21 I agree, but the testcases he have, doesn’t include certificate-data hence the tests passes, (which comply with the default behaviour of client-go) If we use the flag along with having certificate-data unless it is removed manually (which is not good) the behaviour will be overriden.

Alright . So what do you think he should do, should he leave the way it was or he should make the certificate-data. Which is better

yash37158:
I appreciate everyone's review and suggestions. I'm working on the changes and will push them as soon as possible.

yash37158:
@MUzairS15, removing certificate-authority-data is a good approach as Philips said removing may affect the mesheryctl operations.

utils/kubernetes/client.go
yash37158: yeah sure e

yash37158:
Hey @Philip-21 @MUzairS15, I have made the necessary changes suggested by @MUzairS15

utils/kubernetes/client.go
```diff
@@ -1,25 +1,68 @@
...
 		if err != nil {
 			return nil, err
 		}
 
-		if config, err = clientcmd.RESTConfigFromKubeConfig(cfgFile); err == nil {
-			return config, err
+		if err := yaml.Unmarshal(cfgFile, models); err != nil {
+			return nil, err
+		}
+
```
MUzairS15, line 41: Please extract this logic in separate function.

MUzairS15 (changes requested):
@yash37158 Thanks for working on this, going through reviews and making changes. 

Current Behaviour:
1. If `insecure-skip-tls-verify: true/false` and  `certificate-authority-data` is present use the certs. If `certificate-authority-data` not present and `insecure-skip-tls-verify: true` don't use the cert.

Proposes behaviour:
1. If `insecure-skip-tls-verify: true` don't respect the `certificate-authority-data` attempt connection w/o certs.

The proposed fix adds convenience, but because most of the users will be accustomed to what is already present in k8s _The merge of this PR would require a PR explaining this behaviour in details in meshery docs._

utils/kubernetes/client.go
```diff
@@ -13,14 +18,44 @@ import (
...
-		if config, err = clientcmd.RESTConfigFromKubeConfig(cfgFile); err == nil {
-			return config, err
+		models := &models.Kubeconfig{}
+
+    cfgFile, err := processConfig(configfile)
+		if err != nil {
+			return nil, err
+		}
+
+		if err := yaml.Unmarshal(cfgFile, models); err != nil {
```
Aisuko, line 32:
```suggestion
		if err = yaml.Unmarshal(cfgFile, models); err != nil {
```

utils/kubernetes/client.go
```diff
@@ -13,14 +18,44 @@ import (
...
-		cfgFile, err = processConfig(configfile)
 		if err != nil {
 			return nil, err
 		}
 
-		if config, err = clientcmd.RESTConfigFromKubeConfig(cfgFile); err == nil {
-			return config, err
+		models := &models.Kubeconfig{}
+
+    cfgFile, err := processConfig(configfile)
```
Aisuko, line 27:
```suggestion
    cfgFile, err = processConfig(configfile)
```

Aisuko: Using `go fmt` in here should be more better.

yash37158:
@Aisuko Thanks for your review. I have added the changes you asked for

Aisuko:
hi @yash37158 , It looks like we need to sign off https://github.com/meshery/meshkit/pull/373/checks?check_run_id=18097641313

yash37158:
Hey @Aisuko, I already signoff my latest commit. I have added your suggestion from the github itself. below is the screenshot attached. where did I go wrong?
Screenshot 2023-10-28 at 12 20 29 AM

utils/kubernetes/kompose/convert.go
```diff
@@ -193,6 +193,7 @@ func convert(opt kobject.ConvertOptions) error {
 	if err != nil {
 		return err
 	}
+	fmt.Println(komposeObject)
```
Aisuko, line 196: Why print the obj?

Aisuko (changes requested):

utils/kubernetes/kompose/convert.go
yash37158: This komposeObject was throwing an error on linting check as it was initialized and was not used anywhere.

Aisuko: Maybe we can remove it.

yash37158: sure @Aisuko

yash37158:
@Aisuko, Are we good now?

PR #374, opened 2023-09-23
[Schemas] Define all fields in ui:order to prevent crash

Yashsharma1911:
**Description**

This PR fixes #

Recently we get `*` wild card issues in Meshery from RJSF form and error was `You can't use multiple wild cards in ui:order` even if we were using the single wild card and removing it fixes the issue in Meshery and works fine

As a result, it crashes in Meshery cloud RJSF forms and the form and error is `include all field in ui:order` I think this was because of different RJSF versions in both applications, though to prevent that in the future we will make sure to not use `*` wild card and explicitly define all fields in ui:order

FYI wild card helps us not to explicitly define ui order for all fields in form all the time, with the help of wild card we can define the order of some fields and the rest can render after that

**Notes for Reviewers**

### DIFF
--- a/schemas/configuration/uiSchemaApplication.json
+++ b/schemas/configuration/uiSchemaApplication.json
@@ -2,5 +2,5 @@
   "applicationType": {
     "ui:widget": "radio"
   },
-  "ui:order" : ["name", "applicationType"]
+  "ui:order" : ["name", "applicationType", "file", "url"]
 }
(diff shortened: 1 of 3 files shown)


Yashsharma1911:
I'm testing these changes in meshery

Yashsharma1911:
it's working in both cloud and Meshery, merging it and making release

PR #375, opened 2023-09-27
[chore] Improve MeshModels models retrieval

theBeginner86:
**Description**

This PR fixes # https://github.com/meshery/meshery/issues/8929

Fixes:
1. Fix unwanted re-registering of hosts. Earlier same host was being re-registered for each component. So we had about ~4000 entries for ArtifactHub. Now it would be just one.
2. Ensure Models are too registered as an entity in `registries` table. Also, made model compatible with Entity interface{}.
3. Enhanced GetModels handler to do the heavy lifting and return all the necessary info to client instead.

**Notes for Reviewers**

### DIFF
--- a/models/meshmodel/registry/host.go
+++ b/models/meshmodel/registry/host.go
@@ -56,7 +56,7 @@ func createHost(db *database.Handler, h Host) (uuid.UUID, error) {
 
 	// else return the id of the existing host
 	fmt.Println("found id: ", host.ID)
-	return h.ID, nil
+	return host.ID, nil
 }
 
 func (h *Host) AfterFind(tx *gorm.DB) error {
--- a/models/meshmodel/registry/host.go
+++ b/models/meshmodel/registry/host.go
@@ -1,7 +1,6 @@
 package registry
 
 import (
-	"errors"
 	"fmt"
 	"time"
 	"encoding/json"
@@ -60,8 +59,8 @@ func (h *Host) AfterFind(tx *gorm.DB) error {
 		h.IHost = ArtifactHub{}
 	case "kubernetes":
 		h.IHost = Kubernetes{}
-	default:
-		return ErrUnknownHost(errors.New("unable to find compatible host for the component"))
+	default: // do nothing if the host is not pre-unknown. Currently adapters fall into this case.
+		return nil
 	}
 	return nil
 }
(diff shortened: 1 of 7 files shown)


theBeginner86:
I'm unclear of what is the meaning of `hostname` in hosts table. Does it mean name of host or the hostname (the one used in networking)?
I'm asking this because for registrants like ArtifactHub or Kubernetes the value of hostname is `artifact hub` and `kubernetes` respectively but for Meshery Adapters its `localhost` (considering I was running adapter locally).

Isn't this inconsistent?

models/meshmodel/registry/host.go
```diff
@@ -34,8 +59,8 @@ func (h *Host) AfterFind(tx *gorm.DB) error {
 		h.IHost = ArtifactHub{}
 	case "kubernetes":
 		h.IHost = Kubernetes{}
-	default:
-		return ErrUnknownHost(errors.New("unable to find compatible host for the component"))
+	default: // do nothing if the host is not pre-unknown. Currently adapters fall into this case.
```
theBeginner86, line 62: This change relates to the question I asked here: https://github.com/meshery/meshkit/pull/375#issuecomment-1738593411

leecalcote:
> I'm unclear of what is the meaning of `hostname` in hosts table. Does it mean name of host or the hostname (the one used in [networking](https://en.wikipedia.org/wiki/Hostname#:~:text=In%20the%20Internet%2C%20a%20hostname,the%20domain%20name%20wikipedia.org.))? I'm asking this because for registrants like ArtifactHub or Kubernetes the value of hostname is `artifact hub` and `kubernetes` respectively but for Meshery Adapters its `localhost` (considering I was running adapter locally).
> 
> Isn't this inconsistent?

Answered verbally.

leecalcote (approved):

models/meshmodel/registry/registry.go
```diff
@@ -47,6 +49,36 @@ type RegistryManager struct {
 	db *database.Handler //This database handler will be used to perform queries inside the database
 }
 
+func registerModel(db *database.Handler, regID, modelID uuid.UUID) error {
```
theBeginner86, line 52:
```suggestion
// Registers models into registries table.
func registerModel(db *database.Handler, regID, modelID uuid.UUID) error {
```

leecalcote (approved):

PR #376, opened 2023-09-28
Refactor and move `Patterns` constructs.

MUzairS15:
**Description**

This PR 
1. Refactors and moves `Patterns` constructs into meshkit, to ensure any component can use it.
This will also be helpful when we incorporate WASM, as some `Pattern` constructs would be required and importing the meshery server as a package is not ideal.
2. Adds `description` and `icon` attributes as `HelmMetadata`. 
**Notes for Reviewers**


utils/patterns/error.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+func ErrCreatePatternService(err error) error {
+	return errors.New(ErrCreatePatternServiceCode, errors.Alert, []string{"Failed to create design service from Manifest"}, []string{err.Error()}, []string{"Invalid Manifest", "Meshery doesn't identifies the Resource mentioned in the Manifest"}, []string{"Check if all of the meshery adapters are running", "Check if Meshery has successfully identified and registered Kubernetes components"})
+}
+
+func ErrPatternFromCytoscape(err error) error {
+	return errors.New(ErrPatternFromCytoscapeCode, errors.Alert, []string{"Could not create design file from given cytoscape"}, []string{err.Error()}, []string{"Invalid cytoscape body", "Service name is empty for one or more services", "_data does not have correct data"}, []string{"Make sure cytoscape is valid", "Check if valid service name was passed in the request", "Make sure _data field has \"settings\" field"})
```
leecalcote, line 30: @MUzairS15 let's not speak to users in terms of the tech that we use. Please put on your user hat and re-write.

leecalcote (changes requested):

utils/patterns/utils.go
```diff
@@ -0,0 +1,113 @@
...
+	"strings"
+)
+
+// RecursiveCastMapStringInterfaceToMapStringInterface will convert a
+// map[string]interface{} recursively => map[string]interface{}
+func RecursiveCastMapStringInterfaceToMapStringInterface(in map[string]interface{}) map[string]interface{} {
+	res := convertMapInterfaceMapString(in)
+	out, ok := res.(map[string]interface{})
+	if !ok {
+		fmt.Println("failed to cast")
```
leecalcote, line 17: We don't use Println...

utils/patterns/utils.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	res := map[string]interface{}{}
+	if err := json.Unmarshal(byt, &res); err != nil {
+		return map[string]interface{}{}
+	}
+	return res
+}
+
+func GetRandomAlphabetsOfDigit(length int) (s string) {
```
leecalcote, line 105: Function description needed.

utils/utils.go
```diff
@@ -245,3 +245,19 @@ func Contains[G []K, K comparable](slice G, ele K) bool {
 	}
 	return false
 }
+
+func ManifestIsEmpty(manifests []string) bool {
```
leecalcote, line 249: Please add function description.

utils/utils.go
```diff
@@ -245,3 +245,19 @@ func Contains[G []K, K comparable](slice G, ele K) bool {
...
+	for _, m := range manifests {
+		x := strings.TrimSpace(strings.Trim(m, "\n"))
+		if x != "---" && x != "" {
+			return false
+		}
+	}
+	return true
+}
+
```
leecalcote, line 258: Please add function description.

leecalcote (changes requested):

models/events/database.go
```diff
@@ -19,7 +19,6 @@ func (e *Event) BeforeUpdate(tc *gorm.DB) (err error) {
 }
 
 func isEventStatusSupported(event *Event) error {
-
 	if event.Status != Read && event.Status != Unread {
```
leecalcote, line 22: Should this reference a struct/iota/enum from the Eventv2 schema?

utils/kubernetes/apply-helm-chart.go
```diff
@@ -74,10 +74,12 @@ type HelmEntries map[string][]HelmEntryMetadata
 // HelmEntryMetadata is the struct for holding the metadata
 // associated with a helm repositories' entry
 type HelmEntryMetadata struct {
-	APIVersion string `yaml:"apiVersion"`
-	AppVersion string `yaml:"appVersion"`
-	Name       string `yaml:"name"`
-	Version    string `yaml:"version"`
+	APIVersion  string `yaml:"apiVersion"`
```
leecalcote, line 77: Is there anyway for MeshKit to reference the Helm Controller model, using values ComponentDefinitions (granted that we wouldn't support all types of Helm connections immediately)?

utils/patterns/error.go
```diff
@@ -0,0 +1,31 @@
...
+	return errors.New(ErrGetK8sComponentsCode, errors.Alert, []string{"Could not get K8s components for registration"}, []string{err.Error()}, []string{"Invalid kubeconfig", "Filters passed incorrectly in config", "Could not fetch API resources from Kubernetes server"}, []string{"Make sure that the configuration filters passed are in accordance with output from /openapi/v2"})
+}
+
+func ErrParseK8sManifest(err error) error {
+	return errors.New(ErrParseK8sManifestCode, errors.Alert, []string{"Failed to parse manifest into JSON"}, []string{err.Error()}, []string{"Ensure manifests are valid and also use the restricted YAML features like using only \"strings\" in the field names"}, []string{"Ensure YAML is a valid Kubernetes Manifest"})
+}
+
+func ErrCreatePatternService(err error) error {
+	return errors.New(ErrCreatePatternServiceCode, errors.Alert, []string{"Failed to create design service from Manifest"}, []string{err.Error()}, []string{"Invalid Manifest", "Meshery doesn't identifies the Resource mentioned in the Manifest"}, []string{"Check if all of the meshery adapters are running", "Check if Meshery has successfully identified and registered Kubernetes components"})
```
leecalcote, line 26: Capitalize Meshery Adapters.

What type of Manifests are these? Kubernetes? If so, write "Kubernetes manifests".

utils/patterns/pattern.go
```diff
@@ -0,0 +1,698 @@
...
+	//Vars will be used to configure the pattern when it is imported from other patterns.
+	Vars map[string]interface{} `yaml:"vars,omitempty" json:"vars,omitempty"`
+	// PatternID is the moniker use to uniquely identify any given pattern
+	// Convention: SMP-###-v#.#.#
+	PatternID string              `yaml:"patternID,omitempty" json:"patternID,omitempty"`
+	Services  map[string]*Service `yaml:"services,omitempty" json:"services,omitempty"`
+}
+
+// Service represents the services defined within the appfile
+type Service struct {
```
leecalcote, line 158: Outside of backwards compatibility and other changes that would be needed, there is no longer a need for `Service`, is there?

utils/patterns/pattern.go
```diff
@@ -0,0 +1,698 @@
+package patterns
```
leecalcote, line 1: Where is the OpenAPI schema for this resource?  Let's ensure that this package, structs, functions are based directly on that schema.

leecalcote (changes requested):

models/events/database.go
MUzairS15: It is referencing, if you see the schema Status is defined as an enum

utils/kubernetes/apply-helm-chart.go
MUzairS15: This struct represents the index.yml and is used when deploying/fetching the helm charts. Referencing the Helm Controller model wouldn't serve the purpose here.

utils/patterns/pattern.go
MUzairS15: > there is no longer a need for Service

Yes.
The pattern engine will also require changes.

utils/patterns/pattern.go
MUzairS15: Below is the spec for `Meshery Patterns`,  the structs defined here are internal (except DryRun).
https://github.com/meshery/meshery/blob/db401c30f56ff47eae6849bf19aad0c32d8dac2a/server/helpers/swagger.yaml#L2036

utils/kubernetes/apply-helm-chart.go
leecalcote: It would serve that purpose, but also many other purposes. I'm asking if we can reference that component, and only expose a single field.

PR #383, opened 2023-10-09
feat:retrieve registrants and number of models components relationshi…

Jougan-0:
…p and policies it registers w/signoff


models/meshmodel/registry/registry.go
```diff
@@ -201,7 +201,39 @@ func (rm *RegistryManager) RegisterEntity(h Host, en Entity) error {
...
+	var hosts []*Host
+	if limit == -1 {
+		if err := db.Table("hosts").Select("*").Count(&hostCounts).Error; err != nil {
+			return nil, 0, err
+		}
+		limit = int(hostCounts)
+	}
+	if err := db.Table("hosts").
+		Select("hosts.id ,hosts.hostname,hosts.port").
+		Where("hostname LIKE ?", "%"+search+"%").
```
theBeginner86, line 227: what about the case, when `search` is empty?

theBeginner86: Can you please refer how other gorm handlers are defined for models, components, relationships etc?

Example: https://github.com/meshery/meshkit/blob/master/models/meshmodel/registry/registry.go#L232
They use a standard filter to define different parameters for the query.

theBeginner86 (changes requested):

models/meshmodel/registry/registry.go
```diff
@@ -201,7 +201,39 @@ func (rm *RegistryManager) RegisterEntity(h Host, en Entity) error {
 		return nil
 	}
 }
+func (rm *RegistryManager) GetCountForHostAndType(db *gorm.DB, hostID uuid.UUID, typeName string) (int64, error) {
+	var count int64
+	err := db.Table("hosts").
+		Select("hosts.id").
+		Joins("LEFT JOIN registries ON hosts.id = registries.registrant_id").
+		Where("hosts.id = ? AND registries.type LIKE ?", hostID, "%"+typeName+"%").
```
theBeginner86, line 209: Since your function call [here](https://github.com/meshery/meshery/pull/9040/files#diff-53b7da6cb4fd0ed620df83501e2b6b09a0648c5a84e3a9d589c06e02837a7655R1255) is not searching the type then lets not use `LIKE`.
Will you directly use the equality operator?

models/meshmodel/registry/registry.go
Jougan-0: > what about the case, when `search` is empty?

when the search query is empty it works fine I checked it ,but since it can cause confusion will add a case for it too .

> Can you please refer how other gorm handlers are defined for models, components, relationships etc?
> 
> Example: https://github.com/meshery/meshkit/blob/master/models/meshmodel/registry/registry.go#L232 They use a standard filter to define different parameters for the query.

I saw the filters but didn't knew whether I should use it that's why I was waiting for your review 

Will get back to u very soon

models/meshmodel/registry/registry.go
Jougan-0: okay roger that

MUzairS15 (changes requested):

models/meshmodel/core/v1alpha1/host.go
```diff
@@ -0,0 +1,25 @@
+package v1alpha1
+
+import "github.com/google/uuid"
+
+type MesheryHostsDisplay struct {
```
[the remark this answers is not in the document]
Jougan-0: We only need to display some data from host table and summary of models components and relationship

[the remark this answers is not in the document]
Jougan-0: And Retrieving the data in hosts alone was causing issue of non mapping and the values were being displayed individually

theBeginner86 (commented):
I have made the leftover changes and cleanup this PR 👍

theBeginner86 (approved):
Thanks @Jougan-0 for consistent effort in getting this completed.

LGTM!

theBeginner86:
@MUzairS15 looks good to you?

PR #408, opened 2023-11-08
Handle errors while parsing crds.

MUzairS15:
**Description**
Due to malformed CRDs (yaml manifests), the https://github.com/meshery/meshkit/compare/master...MUzairS15:meshkit:MUzairS15/comp-generation/bug-fix?expand=1#diff-372558418b9928d66f2ad2c8891dcafdf9e7062e16ac921e44222a73852d9be9L14
gets into an infinite loop, since the code didn't handle the error correctly.

 
This PR fixes the above bug. We immediately fail and return skipping the process as soon as error is encountered.

**Notes for Reviewers**


Jougan-0 (commented):
The thing is this method was causing some error when I was running it locally we would have to specify the error

leecalcote:
Thanks!

PR #428, opened 2023-12-05
[feat][chores] - Add enchanments for faster data retrieval.

VihasMakwana:
**Description**

This PR fixes [#9494](https://github.com/meshery/meshery/issues/9494).
Add indexes on modelID for faster retrieval.

**Notes for Reviewers**


leecalcote:
@VihasMakwana, thank you for this. Would you be so kind as to propose the associated API changes in [this spreadsheet](https://docs.google.com/spreadsheets/d/1ABJCbfQRi0uN_YoP2kmHZ-lTI4S4QuvRk1_unomKNRE/edit#)?

models/meshmodel/registry/registry.go
```diff
@@ -383,6 +388,24 @@ func (rm *RegistryManager) GetModels(db *database.Handler, f types.Filter) ([]v1
 		model.HostID = host.ID
 		model.HostName = host.Hostname
 		model.DisplayHostName = host.Hostname
+
+		if includeComponents {
+			var components []v1alpha1.ComponentDefinitionDB
+			finder := db.Model(&v1alpha1.ComponentDefinitionDB{}).
+				Select("component_definition_dbs.*").
+				Where("component_definition_dbs.model_id = ?", model.ID)
```
MUzairS15, line 396: With this, the component schema would also be included, which is not required in the registry view.

models/meshmodel/registry/registry.go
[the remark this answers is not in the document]
VihasMakwana: Okay. The reason why my description has "db indexes" is because I've added index tag to some fields, which will necessarily create an index in sqlite db.

> What's the intended affect of this change

When we try to lookup for `components` for a given model, we will perform a sql query.
This can be speed up introducing an index for `modelID` field.
I observed that the latency for SQL was reduced by a whopping ~70%.

This will further help up reduce the latency we're facing in UI rendering.

models/meshmodel/registry/registry.go
VihasMakwana: Hmm. So, you're proposing to introduce only necessary fields, correct?

models/meshmodel/registry/registry.go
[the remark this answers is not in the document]
VihasMakwana: @leecalcote I've updated the PR title to be more meaningful. Thanks

models/meshmodel/registry/registry.go
Yashsharma1911: I think yes, adding a flag can be nice, in case we might want to retrieve other data too.

Yashsharma1911: We usually have `trim` param to avoid sending schema if it is not required checkout [doc](https://docs.meshery.io/reference/rest-apis#api-meshmodels-models-model-components)

MUzairS15: Yes, we do have trim param but for components API not here, since model definitions didn't have schema as such.

@VihasMakwana We want a functionality to selectively retrieve each of the attributes, the fields to retrieve could be passed on as a filter. Though this is something we want but not in this PR. (Something to keep in mind).

Ensuring just the schema is not passed we are good here.

VihasMakwana: Yeah. makes sense

VihasMakwana: I've only requested necessary fields i.e. kind, modelName, apiVersion.
Please re-review @Yashsharma1911 @MUzairS15

models/meshmodel/registry/registry.go
```diff
@@ -383,6 +388,28 @@ func (rm *RegistryManager) GetModels(db *database.Handler, f types.Filter) ([]v1
 		model.HostID = host.ID
 		model.HostName = host.Hostname
 		model.DisplayHostName = host.Hostname
+
+		if includeComponents {
+			var components []v1alpha1.ComponentDefinitionDB
+			finder := db.Model(&v1alpha1.ComponentDefinitionDB{}).
+				Select("component_definition_dbs.kind, component_definition_dbs.display_name, component_definition_dbs.api_version").
```
MUzairS15, line 395:
```suggestion
				Select("component_definition_dbs.kind, component_definition_dbs.display_name, component_definition_dbs.api_version, component_definition_dbs.metadata").
```

MUzairS15 (approved):
Nice work @VihasMakwana .

PR #457, opened 2024-02-06
Unmarshal components metadata when sending with models

Yashsharma1911:
**Description**

This PR fixes #

**Notes for Reviewers**


leecalcote:
Have we discussed this one? Does this one propose a change in the file format over the wire?

MUzairS15 (approved):
Have you verified if other clients are compatible (or updated if required)?

Yashsharma1911:
> Have you verified if other clients are compatible (or updated if required)?

Yes I checked last time, however, I'll re-confirm and merge this

Yashsharma1911:
> Have we discussed this one? Does this one propose a change in the file format over the wire?

Yes Lee this PR is already discussed, This PR make changes to make sure we are sending component metadata in meshmodels/models API in JSON format instead of base64, In our APIs we don't send data in base64 we decided to update this with JSON format however before merging it's necessary to check if there is any client expecting base64 I've checked I didn't find however I'll reconfirm

Yashsharma1911:
Thankyou gents but I'm closing this PR, this issue I fixed from UI itself see PR https://github.com/meshery/meshery/pull/10171, I was giving thought, it would be unnecessary to send all that component information with meshmodels/models API this might can increase latency instead of that it would be nice if UI make sure to get that component information from components API

PR #477, opened 2024-03-14
Count entities

Jougan-0:
**Description**

This PR fixes #

**Notes for Reviewers**


Yashsharma1911 (commented):
@Jougan-0 can you please update the description and title of PR if you can

models/meshmodel/registry/registry.go
```diff
@@ -144,21 +150,28 @@ func (rm *RegistryManager) RegisterEntity(h Host, en Entity) error {
 			CreatedAt:    time.Now(),
 			UpdatedAt:    time.Now(),
 		}
-		return rm.db.Create(&entry).Error
+		err = rm.db.Create(&entry).Error
+		if err != nil {
+			onEntityError(entity, h, err)
```
Yashsharma1911, line 155: @Jougan-0 thanks for raising PR, if you update the title and description it can be more clear to me on what this PR is about, still with considering that you are helping us to make sure we log the errors that we get on performing operations related to registry. Sounds like a awesome idea to me, however can you make this error function more a utility function to be use?

We already have [`utils.Log.Info()`](https://github.com/Yashsharma1911/meshery/blob/d17238391a69a0b88d7abf4980297cbc2c9e5260/mesheryctl/internal/cli/root/registry/generate.go#L97) logic available in mesheryctl which takes any log data you want to put and creates and updates the existing `registry-generate.txt` in your .meshery/logs directory.

I think the utility logger that we have in mesheryctl is more generic in nature and can be reuse as we want, compare to logic like onEntityError which is tightly bound to registry only and can not be reused and not enough generic I say, what do you think?

Lemme know if there any hiccups in between doing, if not I think reusing mesheryctl logic can be much more helpful or move mesheryctl logger logic in meshkit and mesheryctl should use that utility logger from meshkit itself.

Yashsharma1911: @MUzairS15 @RipulHandoo What you guys think?

Jougan-0: > @Jougan-0 thanks for raising PR, if you update the title and description it can be more clear to me on what this PR is about, still with considering that you are helping us to make sure we log the errors that we get on performing operations related to registry. Sounds like a awesome idea to me, however can you make this error function more a utility function to be use?
> 
> We already have [`utils.Log.Info()`](https://github.com/Yashsharma1911/meshery/blob/d17238391a69a0b88d7abf4980297cbc2c9e5260/mesheryctl/internal/cli/root/registry/generate.go#L97) logic available in mesheryctl which takes any log data you want to put and creates and updates the existing `registry-generate.txt` in your .meshery/logs directory.
> 
> I think the utility logger that we have in mesheryctl is more generic in nature and can be reuse as we want, compare to logic like onEntityError which is tightly bound to registry only and can not be reused and not enough generic I say, what do you think?
> 
> Lemme know if there any hiccups in between doing, if not I think reusing mesheryctl logic can be much more helpful or move mesheryctl logger logic in meshkit and mesheryctl should use that utility logger from meshkit itself.

We are documenting the models that are successfully registered and those that are unsuccessfully registered. Thank you for evaluating and you got it correct we are logging the registration. I'll investigate the utilities.Log.Info() and consider ways to make the method more reusable.
P.S.: The data is stored and all entities are logged according to their registrant. 
I'll let you know how things are doing and ping you on Slack if I need assistance.

Jougan-0: > @Jougan-0 thanks for raising PR, if you update the title and description it can be more clear to me on what this PR is about, still with considering that you are helping us to make sure we log the errors that we get on performing operations related to registry. Sounds like a awesome idea to me, however can you make this error function more a utility function to be use?
> 
> We already have [`utils.Log.Info()`](https://github.com/Yashsharma1911/meshery/blob/d17238391a69a0b88d7abf4980297cbc2c9e5260/mesheryctl/internal/cli/root/registry/generate.go#L97) logic available in mesheryctl which takes any log data you want to put and creates and updates the existing `registry-generate.txt` in your .meshery/logs directory.
> 
> I think the utility logger that we have in mesheryctl is more generic in nature and can be reuse as we want, compare to logic like onEntityError which is tightly bound to registry only and can not be reused and not enough generic I say, what do you think?
> 
> Lemme know if there any hiccups in between doing, if not I think reusing mesheryctl logic can be much more helpful or move mesheryctl logger logic in meshkit and mesheryctl should use that utility logger from meshkit itself.

Yash let me know if I am wrong but the utils.log.Info is use to log and store the data depending on the scenario .

The onEntityError is the function that stores in a map all the entites that caused an error and the logging part is done by [here](https://github.com/meshery/meshery/blob/f0087365f8a14388f4a59d54c4ba5ab01ebc1df0/server/meshmodel/helper.go#L259)
. We need the onEntityError to count and store the entites that are not succesfully registered .
Hope this clears the explaination rest I'll comment everywhere i can so that the code can be more readable .
Please correct me if I am wrong .

MUzairS15: @Jougan-0 you are creating a mapping of errors for all entities, but how do they get exposed? `FailedMsgCompute` returns a summary for one of the registrants, I believe. So the component registering the model needs to individually call for all registrants?


Because these registry functions are only concerned with interfering with the Database layer, they should not be concerned with summarising and all other stuff, they should just return an error. Instead, the client invoking the registration should ensure this error handling and eventing.

Also, as much as possible the logic should not be tied to the logic here.

models/meshmodel/registry/registry.go
```diff
@@ -116,23 +119,26 @@ func (rm *RegistryManager) Cleanup() {
 func (rm *RegistryManager) RegisterEntity(h Host, en Entity) error {
 	switch entity := en.(type) {
 	case v1alpha1.ComponentDefinition:
-		isAnnotation, _ := entity.Metadata["isAnnotation"].(bool)
-		if entity.Schema == "" && !isAnnotation { //For components which an empty schema and is not an annotation, exit quietly
+		if entity.Schema == "" { //For components with an empty schema, exit quietly
```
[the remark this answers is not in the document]
Jougan-0: Because I asked a few times whether we should count those entities which are with empty schema lee replied me with [this](https://layer5io.slack.com/archives/CFGG6U10E/p1704404768819809?thread_ts=1704404768.819809&cid=CFGG6U10E) 

I might have misunderstood it.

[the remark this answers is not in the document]
MUzairS15: The actor registering these should be responsible and should maintain how many of the models comps (schema or schema-less) were registered and how many/ which of them failed.

Jougan-0:
https://github.com/meshery/meshkit/pull/485
The current v0.7.20 meshkit release was causing the issue so currently working on v0.7.19

PR #481, opened 2024-03-15
Make extract URL func public

Yashsharma1911:
**Description**

This PR fixes #

The ExtractRepoDetailsFromSourceURL function is a public method that calls the private method internally. This function can be accessed outside of the package, while the private method cannot.

**Notes for Reviewers**


Yashsharma1911:
This is small change and doesn't affect other functions, with that said I'm moving forward here

Yashsharma1911:
Lint failure is not related to this PR

leecalcote (approved):

PR #492, opened 2024-04-12
Fixed Registrant 'Artifact Hub' is inappropriately hardcoded

yash37158:
**Description**

This PR fixes #[10437](https://github.com/meshery/meshery/issues/10437)

**Notes for Reviewers**
As this issue pertained to the UI, upon investigation, it was discovered that the root cause lay within the API call. To address this, I have implemented a fix and pushed the changes.
Please take a moment to review the screenshot below showcasing the updated UI after the fix. Your thoughts and feedback are appreciated.
Before:
Screenshot 2024-04-12 at 9 15 50 PM

After:
Screenshot 2024-04-12 at 9 09 54 PM

### DIFF
--- a/models/meshmodel/registry/registry.go
+++ b/models/meshmodel/registry/registry.go
@@ -266,6 +266,15 @@ func (rm *RegistryManager) GetRegistrants(f *v1alpha1.HostFilter) ([]v1alpha1.Me
 	var response []v1alpha1.MeshModelHostsWithEntitySummary
 
 	for _, r := range result {
+		if r.Hostname == "artifacthub" {
+            r.Hostname = "Artifact Hub"
+            err := rm.db.Model(&Host{}).Where("id = ?", r.HostID).Update("hostname", r.Hostname).Error
+            fmt.Println("Checking Updated hostname:", r.Hostname)
+            if err != nil {
+                // Handle the error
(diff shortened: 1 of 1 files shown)


MUzairS15 (changes requested):
@yash37158  Thanks but this fix will not work.
You are trying to fix the cause by going 1 level deep but you need to go all the way down.

yash37158:
@MUzairS15 Thanks for the feedback. I understand that the current fix may not fully address the issue at hand.
 just wanted to run this by you: I've been thinking about the root cause of the problem, and it seems like the culprit might be where we first save the hostname "artifacthub" to the database.
So, instead of trying to update it each time we fetch it, what if I ensure it's saved correctly from the start? I'm thinking should focus on the code section responsible for saving the hostname, particularly the function that handles the creation or update of hosts.

Am I on the right track with this approach?

MUzairS15:
You are somewhat right. But you shouldn’t change how it’s persisted in the db. Displaying the name in prettified format is the UI concern. 
Instead the way we prettified schema as and when required (as requested by client), similarly we should have for others(hosts..)

yash37158:
@MUzairS15 I have created a New PR [link](https://github.com/meshery/meshery/pull/10747)

PR #494, opened 2024-04-29
Shallow cloning through git clone walker

betelgeuse-code:
**Description**

This PR fixes #

Presently git clone walker doesn't do shallow cloning or depth of 1 cloning which clones the unnecessary repo history too, adding `Depth: 1` in cloneOptions will make sure walker does shallow clone which will save much more memory during cloning.

However, [go-git](https://github.com/go-git/go-git) by default doesn't provide a function to identify the size of the repo or a function to add a limit on the specific size, we might need to add our logic in Walker to do that.

**Notes for Reviewers**

### DIFF
--- a/utils/walker/git.go
+++ b/utils/walker/git.go
@@ -145,6 +145,7 @@ func clonewalk(g *Git) error {
 	cloneOptions := &git.CloneOptions{
 		URL:          fmt.Sprintf("%s/%s/%s", g.baseURL, g.owner, g.repo),
 		SingleBranch: true,
+		Depth:        1,
 	}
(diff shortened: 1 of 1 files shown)


Yashsharma1911:
@MUzairS15 we are doing the disk operation here, instead of `in-memory`

https://github.com/meshery/meshkit/blob/67fc308f510e208c1c7501430c6f645c8064a60f/utils/walker/git.go#L142

leecalcote (approved):
Lgtm

leecalcote (approved):

PR #499, opened 2024-05-29
Error handling for registry

Jougan-0:
**Description**

This PR fixes #

**Notes for Reviewers**


Jougan-0:
https://github.com/meshery/meshery/pull/11065

errors/errors.go
```diff
@@ -102,36 +102,38 @@ func (e *Error) ErrorV2(additionalInfo interface{}) ErrorV2 {
 }
 
 func GetCode(err error) string {
-	if obj := err.(*Error); obj != nil && obj.Code != " " {
-		return obj.Code
+	if obj, ok := err.(*Error); ok {
+		if obj.Code != "" {
```
leecalcote, line 106: Does this work fine for `nil` as well or no?

leecalcote: I don't know. I'm just asking.

models/meshmodel/registry/error.go
```diff
@@ -1,14 +1,39 @@
...
+	ErrRegisteringEntityCode           = "meshkit-11101"
+	ErrUnknownHostInMapCode            = "meshkit-11102"
+	ErrCreatingUserDataDirectoryCode   = "meshkit-11103"
 )
 
 func ErrUnknownHost(err error) error {
 	return errors.New(ErrUnknownHostCode, errors.Alert, []string{"host is not supported"}, []string{err.Error()}, []string{"The component's host is not supported by the version of server you are running"}, []string{"Try upgrading to latest available version"})
 }
+func ErrUnknownHostInMap() error {
+	return errors.New(ErrUnknownHostCode, errors.Alert, []string{"Registrant has no error or it is not supported or unknown."}, nil, []string{"The host registering the entites has no errors with it's entites or is unknown."}, []string{"Validate the registrant name again or check /server/cmd/registery_attempts.json for futher details"})
```
leecalcote, line 23: "Registrant has no error...." Hmm. What?

leecalcote (changes requested):

models/meshmodel/registry/error.go
Jougan-0: The part needs rephrasing did it .
The error is resulted when we access the map of the attempts variable with different registrants and it does not contain that registrant so it shouldn't do anything but vihas mentioned that it might panic so the error is for that. I haven't seen this edge case ever happened so I think this is void but still good to have a error if it happens.

Jougan-0:
While testing the fix of panic error the error code was being shown as None event though it existed. New commit should fix it.

Jougan-0:
Should fix this [issue](https://github.com/meshery/meshery/issues/10875) too.

errors/errors.go
```diff
@@ -102,38 +105,90 @@ func (e *Error) ErrorV2(additionalInfo interface{}) ErrorV2 {
...
+	return code
+}
 func GetSeverity(err error) Severity {
-	if obj := err.(*Error); obj != nil {
-		return obj.Severity
+	var severity Severity
+	defer func() {
+		if r := recover(); r != nil {
+			severity = None
+			fmt.Println("Recovered from panic:", r)
```
leecalcote, line 127: I don't know that this is necessary and, at first glance, it seems redundant or not insightful.

Jougan-0: Thanks for mentioning it was redundant, by mistake I pushed defer on top the assertion logic .
Reason behind using anonymous function is the assertion logic, which I used was giving ```NONE``` on meshkit errors that existed when I released my local meshkit release and used it in go.mod but working fine for local meshkit so new logic is that if we get a panic rather than panic and shutting down the process we catch the panic and returns old value.

leecalcote (approved):

leecalcote:
@Jougan-0 will you reconcile the two different error codes for `ErrRegisterEntityCode`?

PR #502, opened 2024-06-04
[Helm] Update DryRun to include Kubernetes version.

MUzairS15:
**Description**

This PR fixes #
1. When performing the dry run using the helm client, `kubernetes` version was not specified leading to `v1.20.0` being used for rendering the templates and converting them to manifests. (static analysis).
This caused an issue when trying to import charts that depend on some specific (or required) Kubernetes version. For eg: In the case of `KEDA` helm chart is `v1.23.0`.

2. The location of `crds` that are bundled inside the charts is usually at top level `crds` folder. (This is also the location expected by the helm client). In cases where the CRDs are included inside some other location or inside `templates` folder, at the time of model generation no CRDs were detected doesn't matter whether the registrant is ArtifactHub/GitHub as we use the helm client `CRDObjects` function.
The logic has been updated to dry run the chart and then extract CRDs.

**Notes for Reviewers**.

### DIFF
--- a/generators/github/git_repo.go
+++ b/generators/github/git_repo.go
@@ -3,14 +3,15 @@ package github
 import (
 	"bufio"
 	"fmt"
-	"github.com/layer5io/meshkit/generators/models"
-	"github.com/layer5io/meshkit/utils"
-	"github.com/layer5io/meshkit/utils/helm"
-	"github.com/layer5io/meshkit/utils/walker"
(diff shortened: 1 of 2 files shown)


leecalcote (commented):
Good sleuthing and great write up. 

Once Meshery's logic has been updated to source the latest supported model version of the target platform, we're good to go here.

PR #503, opened 2024-06-05
Error for yaml unmarshall w/signoff

Jougan-0:
**Description**

This PR fixes 
***BEFORE***
image

***AFTER***
Screenshot from 2024-06-05 19-25-01

**Notes for Reviewers**
New meshkit error for Yaml unmarshal.

### DIFF
--- a/utils/manifests/error.go
+++ b/utils/manifests/error.go
@@ -15,14 +15,14 @@ const (
 	ErrAbsentFilterCode          = "meshkit-11238"
 	ErrCreatingDirectoryCode     = "meshkit-11239"
 	ErrGetResourceIdentifierCode = "meshkit-11240"
-	ErrUnmarshalSyntaxCode       = "meshkit-11241"
+	ErrYamlUnmarshalSyntaxCode   = "replace_me"
 )
 
 func ErrGetResourceIdentifier(err error) error {
 	return errors.New(ErrGetResourceIdentifierCode, errors.Alert, []string{"Error extracting the resource identifier name"}, []string{err.Error()}, []string{"Could not extract the value with the given filter configuration"}, []string{"Make sure to input a valid manifest", "Make sure to provide the right filter configurations", "Make sure the filters are appropriate for the given manifest"})
 }
-func ErrUnmarshalSyntax(err string, packageName string) error {
-	return errors.New(ErrUnmarshalSyntaxCode, errors.Alert, []string{"Unmarshal syntax error."}, []string{fmt.Sprintf("helm chart of %s has syntax error(s): \n%s", packageName, err)}, []string{"Invalid object format"}, []string{"Make sure to input a valid YAML object"})
+func ErrYamlUnmarshalSyntax(err string, packageName string) error {
+	return errors.New(ErrYamlUnmarshalSyntaxCode, errors.Alert, []string{"Unmarshal syntax error."}, []string{fmt.Sprintf("helm chart of %s has syntax error(s): \n%s", packageName, err)}, []string{"Invalid object format"}, []string{"Make sure to input a valid YAML object"})
 }
 func ErrGetCrdNames(err error) error {
 	return errors.New(ErrGetCrdNamesCode, errors.Alert, []string{"Error getting crd names"}, []string{err.Error()}, []string{"Could not execute kubeopenapi-jsonschema correctly"}, []string{"Make sure the binary is valid and correct", "Make sure the filter passed is correct"})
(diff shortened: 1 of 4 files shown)


utils/manifests/getComponents.go
```diff
@@ -49,7 +49,10 @@ func GetCrdsFromHelm(url string) ([]string, error) {
 			if err == io.EOF {
 				break
 			}
-			return nil, err
+			errStr := err.Error()
+			//The first line is "yaml unmarshall error:".
+			errStr = strings.TrimPrefix(errStr, strings.SplitN(errStr, "\n", 2)[0]+"\n")
```
MUzairS15, line 54: Why this is needed?

utils/manifests/error.go
```diff
@@ -11,12 +15,15 @@ const (
 	ErrAbsentFilterCode          = "meshkit-11238"
 	ErrCreatingDirectoryCode     = "meshkit-11239"
 	ErrGetResourceIdentifierCode = "meshkit-11240"
+	ErrUnmarshalSyntaxCode       = "meshkit-11241"
```
MUzairS15, line 18: Read the docs for "How to write mesh kit error codes"

utils/manifests/getComponents.go
Jougan-0: When the error is returned yaml decoder
it has a line ```yaml: unmarshal errors:
  line 6921: mapping key "apiVersion" already defined at line 6427
  line 6922: mapping key "kind" already defined at line 6428
  line 6923: mapping key "metadata" already defined at line 6429
  line 6928: mapping key "spec" already defined at line 6436
  line 7723: mapping key "status" already defined at line 6914```
  we dont need the yaml:unmarshal errors: because we defined our own  meshkit error we only require the line number(s).

MUzairS15: Wil you add a comment so that it's clear in future what this piece of code is doing?

utils/manifests/error.go
```diff
@@ -11,12 +15,15 @@ const (
...
 	ErrGetResourceIdentifierCode = "meshkit-11240"
+	ErrYamlUnmarshalSyntaxCode   = "replace_me"
 )
 
 func ErrGetResourceIdentifier(err error) error {
 	return errors.New(ErrGetResourceIdentifierCode, errors.Alert, []string{"Error extracting the resource identifier name"}, []string{err.Error()}, []string{"Could not extract the value with the given filter configuration"}, []string{"Make sure to input a valid manifest", "Make sure to provide the right filter configurations", "Make sure the filters are appropriate for the given manifest"})
 }
-
+func ErrYamlUnmarshalSyntax(err string, packageName string) error {
+	return errors.New(ErrYamlUnmarshalSyntaxCode, errors.Alert, []string{"Unmarshal syntax error."}, []string{fmt.Sprintf("helm chart of %s has syntax error(s): \n%s", packageName, err)}, []string{"Invalid object format"}, []string{"Make sure to input a valid YAML object"})
```
MUzairS15, line 25: Why are you being specific about the helm chart in this general yaml unmarshal?

utils/manifests/getComponents.go
MUzairS15: Now I think @Jougan-0 does it affect anything? Having that line is no harm. 
Instead, it clears confusion IMO for eg: if error is `line 4545 duplicate field xcv` if we don't have the link it is confusing from where the error is originating is it from validation/marshal or where?
Also, this string manipulation is not worth the benefit.

MUzairS15: Including the model name is helpful, that's ok.

MUzairS15 (changes requested):

utils/manifests/getComponents.go
MUzairS15: so will you remove this logic and only keep adding pkg name in error

Jougan-0: Yes.

utils/manifests/error.go
Jougan-0: I believe this error is no longer necessary since we have resolved the issues with the current model and error naming. Now, errors for a model are displayed directly under its name, making this pull request redundant.

Regarding the specific file mentioned by Lee, @MUzairS15 , I am unable to find a way to implement it. This is because our manifest contains all the files, and if there is an issue with the metadata, the name is not retrieved.

MUzairS15:
can we close this PR, then @Jougan-0

PR #512, opened 2024-06-17
Supplemental: update to get the correct version from URL

Kevin222004:
**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ cmd/errorutil/errorutil
 **errorutil_errors_export.json
 *.DS_Store
 coverage.txt
+.idea
--- a/generators/github/url_test.go
+++ b/generators/github/url_test.go
@@ -1,0 +1,21 @@
+package github
+
+import (
+	"net/url"
+	"testing"
+)
+
+func TestURL_GetVersion(t *testing.T) {
+	UrlString := "https://github.com/kubernetes-sigs/metrics-server"
+	parsedUrl, _ := url.Parse(UrlString)
+	testData := NewDownloaderForScheme("https", parsedUrl, "metrics-server")
+	output, _ := testData.GetContent()
+
+	if output == nil {
+		t.Fatalf("Expected non-nil data, got nil")
+	}
+
+	if version := output.GetVersion(); version != "metrics-server-helm-chart-3.12.1" {
+		t.Errorf("Expected version metrics-server-helm-chart-3.12.1, got %s", version)
+	}
+}
--- a/generators/github/git_repo_test.go
+++ b/generators/github/git_repo_test.go
@@ -1,0 +1,21 @@
+package github
+
+import (
+	"net/url"
+	"testing"
+)
+
+func TestGetVersion(t *testing.T) {
+	gitProtoString := "git://github.com/kubernetes-sigs/metrics-server/master/charts"
+	parsedUrl, _ := url.Parse(gitProtoString)
+	testData := NewDownloaderForScheme("git", parsedUrl, "metrics-server")
+	output, _ := testData.GetContent()
+
+	if output == nil {
+		t.Fatalf("Expected non-nil data, got nil")
+	}
+
+	if version := output.GetVersion(); version != "metrics-server-helm-chart-3.12.1" {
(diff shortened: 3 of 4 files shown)


Kevin222004:
@MUzairS15 as you mentioned is slack chat I have removed the version.

But may be I am not able to explain you this scenario. please take a look at the test case in both the test code will give you the correct version only if the URL is in form. `https://github.com/kubernetes-sigs/metrics-server/metrics-server-helm-chart-3.12.1` 
and this needs to get updated

generators/github/url_test.go
```diff
@@ -0,0 +1,36 @@
...
+	parsedUrl, _ := url.Parse(UrlString)
+	testData := NewDownloaderForScheme("https", parsedUrl, "metrics-server")
+	output, _ := testData.GetContent()
+
+	if output == nil {
+		t.Fatalf("Expected non-nil data, got nil")
+	}
+
+	if version := output.GetVersion(); version != "metrics-server" {
+		t.Errorf("Expected version metrics-server, got %s", version)
```
Kevin222004, line 34: This is not expected correct version is `metrics-server-helm-chart-3.12.1`

If I am missing anything here please let me know or else tell me I will update the code base to get the version in this PR itself.

leecalcote:
Thanks, @Kevin222004 👍

.gitignore
```diff
@@ -23,3 +23,4 @@ cmd/errorutil/errorutil
 **errorutil_errors_export.json
 *.DS_Store
 coverage.txt
+.idea
```
Jougan-0, line 26: Why this??

generators/github/url_test.go
Jougan-0: What is the final goal of this test? 
Check if the version is retrieved but the issue which I see is still we checking with ``` metrics-server-helm-chart-3.12.1``` shouldn't it be just ``` 3.12.1 ```
See here in the logs it still shows ```  metrics-server-helm-chart-3.12.1 ``` instead of just showing the version ``` 3.12.1 ```

The main problem was we were showing 
``` Generated 0 components for model [metrics-server] metrics-server ```

where it should have been

```Generated 0 components for model [metrics-server] 3.12.1 ``` .
The test case should also follow these norms imo.

Here is the current log that is being generated in the recent [workflow run ](https://github.com/meshery/meshery/actions/workflows/model-generator.yml)

The issue still persist even though we changed the link it still shows. 
``` Generated 0 components for model [metrics-server] metrics-server-helm-chart-3.12.1```
image

Kevin222004: The final goal of the test should be to get the correct version of the URL provided.

The new link that is been changed is of type git protocol.  which use this method to get all the version 
https://github.com/meshery/meshkit/blob/efd59b7233304322ba2ee2ee013b127f511cfecc/utils/utils.go#L205-L207
please checkout https://github.com/kubernetes-sigs/metrics-server/tags in their releases the tag that has been ` metrics-server-helm-chart-3.12.1` that's why we are getting the version like this but if I talk about the URL with `https` protocol the implementation is wrong. i don't have much idea about releases but if you want that we need only the latest version as they mentioned https://github.com/kubernetes-sigs/metrics-server/releases#:~:text=v0.7.1-,Latest,-Installation I will find some way to get that. 
 
until then I have updated repo so that if you use URL you get the latest verison

.gitignore
Kevin222004: it is like `.vscode/`  for jetbrains ide

generators/github/url_test.go
Jougan-0: Please update and ping me if we need to update the github url in the sheel mention I'll update and let's see the logs.

Kevin222004: Yup lets update it and see what logs are getting generated know with the URL `https://github.com/kubernetes-sigs/metrics-server`

leecalcote:
Re-running build workflows...

Kevin222004:
@leecalcote @Jougan-0 If you are good with the changes can you merge it :)

Kevin222004:
@Jougan-0 let's finish this pr.

PR #529, opened 2024-07-06
remove version from github url w/signoff

Jougan-0:
**Description**

This PR fixes #

**Notes for Reviewers**


MUzairS15:
@Jougan-0 thanks for this. But afterwards I realised when we point towards a particular file we are not aware of the software/project version to pack the generated model. 
hence we might need this.

If you can propose a new format such as version to be specified as query param that will be helpful.
It will also help to identify and error out explicitly specifying the reason, if version has not been provided.

Jougan-0:
Okay closing this then.

PR #542, opened 2024-07-20
Source class description from schema

shubham251972:
**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/models/catalog/v1alpha1/catalog.go
+++ b/models/catalog/v1alpha1/catalog.go
@@ -5,12 +5,13 @@ import (
 	"encoding/json"
 
 	"github.com/layer5io/meshkit/utils"
+	"github.com/meshery/schemas"
 )
 
 // CatalogData defines model for catalog_data.
 type CatalogData struct {
 	ContentClass ContentClass `json:"content_class,omitempty"`
-	//Tracks the specific content version that has been made available in the Catalog
+	// Tracks the specific content version that has been made available in the Catalog
 	PublishedVersion string `json:"published_version"`
 
 	// Compatibility A list of technologies included in or implicated by this design; a list of relevant technology tags.
@@ -23,7 +24,7 @@ type CatalogData struct {
 	PatternInfo string `json:"pattern_info"`
 
 	// Contains reference to the dark and light mode snapshots of the catalog.
-	SnapshotURL []string `json:"imageURL,omitempty"` // this will require updating exisitng catalog data as well. updated the json tag to match previous key name, so changes will not be required in exisitng catgalogs
+	SnapshotURL []string `json:"imageURL,omitempty"` // this will require updating existing catalog data as well. updated the json tag to match previous key name, so changes will not be required in existing catalogs
 
 	// Type Categorization of the type of design or operational flow depicted in this design.
 	Type CatalogDataType `json:"type"`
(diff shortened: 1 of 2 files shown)


shubham251972:
This PR depends on meshery/schemas new version, don't merge this until we release new version of schemas repo

shubham251972:
Lint failure will get pass as we will make Schemas repo release

models/catalog/v1alpha1/catalog.go
```diff
@@ -80,33 +81,59 @@ const (
...
-	case Official:
-		return "official"
-	case Verified:
-		return "verified"
-	case Community:
-		fallthrough
-	default:
-		return "community"
+// Load and extract descriptions and enum values from schema
+func loadSchemaDescriptions() (map[string]ContentClassObj, error) {
```
leecalcote, line 85: Meshery Schemas are to be imported as released golang packages. 

I'm trying to understand why this approach is necessary.

leecalcote (changes requested):

models/catalog/v1alpha1/catalog.go
Yashsharma1911: Yes we are importing them as golang package 

https://github.com/meshery/meshkit/pull/542/files#diff-6aee5a74826065e2f1c176431a0d3f7274d9baac2a66d75d97cf1efad6c452aaR8

This part extracts the class description from the schema which we get from released schema golang package, we can change this func name to a better descriptive name like `extractClassAndDescriptions `

shubham251972: Yes we are importing them as golang package

https://github.com/meshery/meshkit/pull/542/files#diff-6aee5a74826065e2f1c176431a0d3f7274d9baac2a66d75d97cf1efad6c452aaR8

This part extracts the class description from the schema which we get from released schema golang package, we can change this func name to a better descriptive name like extractClassAndDescriptions

leecalcote: Thanks. I'm sorry to say that I still don't understand the need for a function specifically for this property and not either a general utility function to retrieve property values from a separate, associated property.

Perhaps, we shouldn't have descriptions in an array to match an enum in a separate property, but use an approach like:

```
{
    "type": "string",
    "oneOf": [
        { "const": "community", "description": "description of the class." },
        { "const": "verified", "description": "another description of the class" },
        { "const": "official", "description": "Best" },
    ]
}
```

leecalcote: Enhancement proposed - https://github.com/meshery/schemas/issues/109

leecalcote: @shubham251972 will you be taking care of this?

shubham251972: JSON schema updated with oneOf - https://github.com/meshery/schemas/pull/126/files

Updated meshkit code to take benefit of `oneOf` and return classes

shubham251972:
Lint is failing because of other PR, this PR depends on schemas repo new release

shubham251972:
> Lint is failing because of other PR, this PR depends on schemas repo new release

Schema is bumped, good to merge this PR

Yashsharma1911:
Shubham test failing, Uhh ho! looks like validator test failing because changes made in schemas v0.7.11 version, @MUzairS15 will this schema bump will cause issues in validator?

MUzairS15 (commented):
@Yashsharma1911 There’s an existing functionality to retrieve schema for resources, can you try using that instead?

Yashsharma1911:
> @Yashsharma1911 There’s an existing functionality to retrieve schema for resources, can you try using that instead?

I tried but I wasn't able to find it, if you can share 😅

leecalcote:
@Yashsharma1911 can you help finish this off?

PR #569, opened 2024-08-19
fix: fetch styles from db w/signoff

Jougan-0:
**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/models/meshmodel/registry/registry.go
+++ b/models/meshmodel/registry/registry.go
@@ -197,10 +197,6 @@ func (rm *RegistryManager) GetEntities(f entity.Filter) ([]entity.Entity, int64,
 	return f.Get(rm.db)
 }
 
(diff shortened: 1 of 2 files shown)


models/meshmodel/registry/registry.go
```diff
@@ -197,10 +197,6 @@ func (rm *RegistryManager) GetEntities(f entity.Filter) ([]entity.Entity, int64,
 	return f.Get(rm.db)
 }
 
-func (rm *RegistryManager) GetEntityById(f entity.Filter) (entity.Entity, error) {
```
Jougan-0, line 200: Used for export We can use the existing function with filter so removed it.

MUzairS15 (approved):

MUzairS15:
To confirm the issue was only occurring when return components param was set in the models API, styles were not returned.

PR #584, opened 2024-08-29
update rjsf gird schema for upload method left side

amitamrutiya:
**Description**

This PR fixes [this discussion](https://layer5io.slack.com/archives/G01BD9056TD/p1724348283333319?thread_ts=1724348174.542299&cid=G01BD9056TD)

**Notes for Reviewers**
Screenshot from 2024-08-29 10-27-02

### DIFF
--- a/schemas/configuration/designImport.json
+++ b/schemas/configuration/designImport.json
@@ -41,7 +41,7 @@
               "URL Import"
             ],
             "default": "URL Import",
-            "x-rjsf-grid-area": "12",
+            "x-rjsf-grid-area": "3",
             "description": "Choose the method you prefer to upload your design file. Select 'File Upload' if you have the file on your local system, or 'URL Import' if you have the file hosted online."
           }
         },


sudhanshutech (changes requested):
@amitamrutiya we are using schema from sistent so fixing here will not do the fix. You have fto make the change in sistent

amitamrutiya:
Yes @sudhanshutech, you are right.

The schema used on this page, https://playground.meshery.io/configuration/designs, is from Sistent.

However, the schema used on the MeshMap page import design is from this schema. I have updated the import model schema accordingly.

leecalcote (approved):
We're in need of consolidating UI Schemas to a single location.

Yashsharma1911:
> Yes @sudhanshutech, you are right.
> 
> The schema used on this page, https://playground.meshery.io/configuration/designs, is from Sistent.
> 
> However, the schema used on the MeshMap page import design is from this schema. I have updated the import model schema accordingly.

@amitamrutiya This is a regression issue, why is this happening now? have we made any changes in RJSF core components?

This schema change should not be needed

amitamrutiya:
Yeah, @Yashsharma1911  you are right. I am not aware that this is a regression issue. Thanks for pointing out. There are definitely some component changes to make this happen.

PR #595, opened 2024-09-27
Update schema for import

Jougan-0:
**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/schemas/configuration/generate.json
+++ b/schemas/configuration/generate.json
@@ -1,0 +1,223 @@
+{
+    "type": "object",
+    "properties": {
+      "uploadType": {
+        "title": "Upload method",
+        "enum": ["URL Import", "CSV Import"],
+        "default": "Select the Upload Method",
+        "x-rjsf-grid-area": "6", 
+        "description": "Choose the method you prefer to upload your model file. Select 'File Upload' if you have the file on your local system or 'URL Import' if you have the file hosted online."
+      },
+      "register": {
+        "type": "boolean",
+        "title": "Register",
+        "default": true,
+        "description": "Indicate whether to register this model.",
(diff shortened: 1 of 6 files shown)


schemas/configuration/generate.json
```diff
@@ -0,0 +1,223 @@
...
+                    "Machine Learning",
+                    "Observability and Analysis",
+                    "Orchestration & Management",
+                    "Platform",
+                    "Provisioning",
+                    "Runtime",
+                    "Security & Compliance",
+                    "Serverless",
+                    "Tools",
+                    "Uncategorized"
```
MUzairS15, line 104: These should be fetched dynamically, look for the IMport design model (source type field) for reference on dynamically fetching.

schemas/configuration/generate.json
```diff
@@ -0,0 +1,223 @@
...
+                  ]
+                },
+                "subCategory": {
+                  "type": "string",
+                  "enum": [
```
MUzairS15, line 109: same here

MUzairS15 (changes requested):

schemas/configuration/generate.json
Jougan-0: @MUzairS15 We used enum here to :thinking: 
https://github.com/meshery/meshkit/blob/fe3024a1caa43e6e9b281c326dac83bb93f93df7/schemas/configuration/designImport.json#L13

MUzairS15: Then it must be some other schema, check catalog publish

Jougan-0: I'll make them dynamically after today's event.

PR #622, opened 2024-12-12
fix operator deployment

Alero-Awani:
**Description**

This PR fixes ##12706

**Notes for Reviewers**


sangramrath:
@Alero-Awani videos were clear and demonstrated that both approaches work.
@leecalcote this also validates that multiple k8s clusters can be added and managed by Meshery.

sangramrath (commented):
Looks goods

leecalcote:
@souvikinator @aabidsofi19, please be aware of this.

leecalcote (commented):
The certfile and keyfile are written to the filesystem using the os.CreateTemp function in the setDataAndReturnFileHandler function. This function creates a temporary file and writes the provided data to it.

There is no explicit check for free space or filesystem permissions in the provided code. The os.CreateTemp function will fail if there is insufficient space or if the process lacks the necessary permissions to write to the temporary directory.

The os.CreateTemp function ensures that a unique temporary file is created, thus preventing overwrites of existing files. The function generates a unique filename with a random component to avoid conflicts with existing files.

leecalcote (commented):
@Alero-Awani, in the event that a user doesn't flatten their kubeconfig, and therefore, ensure that the necessary certificates are included in the contents of their kubeconfig file, are we still following those filepaths and attempting to retrieve the certs?

Ideally, `mesheryctl` does this as well as Meshery Server. The same kubeconfig certificate handling logic should be reused when applying Helm charts for Meshery Operator and any other Kubernetes connection creation flows.

PR #647, opened 2025-01-21
Fix for getMeshSyncStatus

palSagnik:
**Description**

This PR fixes [Meshery Operator Controller checks failing](https://github.com/meshery/meshery-operator/issues/584)

**Notes for Reviewers**

### DIFF
--- a/models/controllers/meshery_operator.go
+++ b/models/controllers/meshery_operator.go
@@ -8,6 +8,7 @@ import (
 	mesherykube "github.com/layer5io/meshkit/utils/kubernetes"
 	kubeerror "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/kubectl/pkg/polymorphichelpers"
 )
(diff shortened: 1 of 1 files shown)


palSagnik:
The frontend logic for MeshSync Ad Hoc Connectivity Test was checking for `Connected` string but the backend was returning `Deployed` or `Deploying` which are not the same.

- ✅ MeshSync chip
- :x: Meshery Broker chip

Will fix in next PR as it needs more trouble shooting

leecalcote:
> The frontend logic for MeshSync Ad Hoc Connectivity Test was checking for `Connected` string but the backend was returning `Deployed` or `Deploying` which are not the same.
> 
> * ✅ MeshSync chip
> * ❌ Meshery Broker chip
> 
> Will fix in next PR as it needs more trouble shooting

Is it the frontend that needs to change? Verify against the Meshery Docs page on troubleshooting these connections, please.

palSagnik:
According to the troubleshooting specs document, the front-end is using the correct terms, there needs to be a change in the backend to fix the ad-hoc connectivity test.

leecalcote:
> According to the troubleshooting specs document, the front-end is using the correct terms, there needs to be a change in the backend to fix the ad-hoc connectivity test.

That may well be true. Thank you for working on this. I need more compelling justification for us to proceed here, however. Research their meanings and differences. This PR eradicates "DEPLOYED" from ever being used? See the Troubleshooting Toolkit google doc.

palSagnik:
After today's call I have understood what you mean, Lee. I will do the required research.

PR #688, opened 2025-02-23
Add registry utility functions

riyaa14:
**Description**

This PR fixes [#13668](https://github.com/meshery/meshery/issues/13668)

**Notes for Reviewers**
#### Files moved completely from `mesheryctl/pkg/utils` to `meshery/meshkit`:
1. model.go
2. relationship.go
3. component.go
4. spreadsheet_update.go

#### New files created in meshkit
5. helper.go : some functions from mesheryctl/pkg/utils/formatter.go and mesheryctl/pkg/utilsregistry.go 
6. error.go : some specific error codes from mesheryctl/pkg/utils/formatter.go 

### DIFF
--- a/files/error.go
+++ b/files/error.go
@@ -10,6 +10,18 @@ import (
 )
 
 var (
+	ErrFileReadCode = "meshkit-11304"
+)
+
+func ErrFileRead(err error) error {
+	return errors.New(ErrFileReadCode, errors.Alert,
+		[]string{"File read error"},
+		[]string{err.Error()},
+		[]string{"The provided file is not present or has an invalid path."},
+		[]string{"To proceed, provide a valid file path with a valid file."})
+}
+
+var (
 	// Error code
 	ErrUnsupportedExtensionCode                    = "meshkit-11282"
 	ErrUnsupportedExtensionForOperationCode        = "meshkit-11283"
--- a/registry/component.go
+++ b/registry/component.go
@@ -1,0 +1,447 @@
+package registry
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/layer5io/meshkit/encoding"
+	"github.com/layer5io/meshkit/models/meshmodel/entity"
+	"github.com/layer5io/meshkit/utils"
+	"github.com/layer5io/meshkit/utils/csv"
+	"github.com/layer5io/meshkit/utils/manifests"
+	"github.com/meshery/schemas/models/v1alpha1/capability"
+	schmeaVersion "github.com/meshery/schemas/models/v1beta1"
+	"github.com/meshery/schemas/models/v1beta1/component"
+)
+
+const (
+	SVG_WIDTH  = 20
+	SVG_HEIGHT = 20
+)
+
+type ComponentCSV struct {
+	Registrant         string `json:"registrant" csv:"registrant"`
+	Model              string `json:"model" csv:"model"`
+	Component          string `json:"component" csv:"component"`
+	Category           string `json:"category" csv:"category"`
+	SubCategory        string `json:"subCategory" csv:"subCategory"`
+	Description        string `json:"description" csv:"description"`
+	Shape              string `json:"shape" csv:"shape"`
+	PrimaryColor       string `json:"primaryColor" csv:"primaryColor"`
+	SecondaryColor     string `json:"secondaryColor" csv:"secondaryColor"`
+	SVGColor           string `json:"svgColor" csv:"svgColor"`
+	SVGWhite           string `json:"svgWhite" csv:"svgWhite"`
+	SVGComplete        string `json:"svgComplete" csv:"svgComplete"`
+	Schema             string `json:"schema" csv:"schema"`
+	Docs               string `json:"docs" csv:"docs"`
+	StyleOverrides     string `json:"styleOverrides" csv:"styleOverrides"`
+	Styles             string `json:"styles" csv:"styles"`
+	ShapePolygonPoints string `json:"shapePolygonPoints" csv:"shapePolygonPoints"`
+	DefaultData        string `json:"defaultData" csv:"defaultData"`
+	Capabilities       string `json:"capabilities" csv:"capabilities"`
+	LogoURL            string `json:"logoURL" csv:"logoURL"`
+	Genealogy          string `json:"genealogy" csv:"genealogy"`
+	IsAnnotation       string `json:"isAnnotation" csv:"isAnnotation"`
+	Version            string `json:"version" csv:"version"`
+
+	ModelDisplayName string `json:"modelDisplayName" csv:"-"`
+
+	Status string `json:"status" csv:"status"`
+}
+
+// The Component Definition generated assumes or is only for components which have registrant as "meshery"
+func (c *ComponentCSV) CreateComponentDefinition(isModelPublished bool, defVersion string) (component.ComponentDefinition, error) {
+	status := entity.Enabled
+	if c.Status != "" {
+		if utils.ReplaceSpacesAndConvertToLowercase(c.Status) == "false" {
+			status = entity.Ignored
+		}
+	}
+	componentDefinition := &component.ComponentDefinition{
+		SchemaVersion: schmeaVersion.ComponentSchemaVersion,
+		DisplayName:   c.Component,
+		Format:        "JSON",
+		Version:       defVersion,
+		Metadata: component.ComponentDefinition_Metadata{
+			Published: isModelPublished,
+		},
+		Status: (*component.ComponentDefinitionStatus)(&status),
+		Component: component.Component{
+			Kind:    c.Component,
+			Schema:  c.Schema,
+			Version: c.Version,
+		},
+	}
+	if c.Description != "" {
+		componentDefinition.Description = c.Description
+	}
+	err := c.UpdateCompDefinition(componentDefinition)
+	return *componentDefinition, err
+}
+
+var compMetadataValues = []string{
+	"genealogy", "isAnnotation", "styleOverrides",
+}
+var compStyleValues = []string{
+	"primaryColor", "secondaryColor", "svgColor", "svgWhite", "svgComplete", "shape",
+}
+
+func (c *ComponentCSV) UpdateCompDefinition(compDef *component.ComponentDefinition) error {
+	status := entity.Enabled
+	if c.Status != "" {
+		if utils.ReplaceSpacesAndConvertToLowercase(c.Status) == "false" {
+			status = entity.Ignored
+		}
+	}
+	compDef.Status = (*component.ComponentDefinitionStatus)(&status)
+	var existingAddditionalProperties map[string]interface{}
+	if c.Description != "" {
+		compDef.Description = c.Description
+	}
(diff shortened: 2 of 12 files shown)


registry/component.go
```diff
@@ -0,0 +1,447 @@
...
+	schmeaVersion "github.com/meshery/schemas/models/v1beta1"
+	"github.com/meshery/schemas/models/v1beta1/component"
+)
+
+const (
+	SVG_WIDTH  = 20
+	SVG_HEIGHT = 20
+)
+
+type ComponentCSV struct {
```
leecalcote, line 25: @riyaa14 this generated from schema using the openapi tool?

leecalcote (commented):
Thank you for placing these files under "registry" folder.

registry/component.go
riyaa14: I didn't generate it but rather this file was copied completely from mesheryctl utils and shifted here in meshkit/registry

registry/spreadsheet_update.go
```diff
@@ -0,0 +1,349 @@
...
+
+	return modelList
+}
+
+// Verifies if the component entry already exist in the spreadsheet, otherwise updates the spreadshhet to include new component entry.
+func VerifyandUpdateSpreadsheet(cred string, wg *sync.WaitGroup, srv *sheets.Service, spreadsheetUpdateChan chan SpreadsheetData, sheetId, modelCSVFilePath, componentCSVFilePath string) {
+	defer wg.Done()
+
+	entriesToBeAddedInCompSheet := []*ComponentCSV{}
+	entriesToBeAddedInModelSheet := []*ModelCSV{}
```
lekaf974, line 96: Moving at top of for loop will allow to remove the needs of `Reset the list`

registry/spreadsheet_update.go
```diff
@@ -0,0 +1,349 @@
...
+			Log.Error(err)
+		}
+	}
+
+	if len(entriesToBeAddedInCompSheet) > 0 {
+		err := updateComponentsSheet(srv, cred, sheetId, entriesToBeAddedInCompSheet, componentCSVFilePath)
+		if err != nil {
+			Log.Error(err)
+		}
+		return
```
lekaf974, line 160: This looks not useful here

registry/spreadsheet_update.go
```diff
@@ -0,0 +1,349 @@
...
+			if err != nil {
+				Log.Error(err)
+			}
+		}
+	}
+
+	if len(entriesToBeAddedInModelSheet) > 0 {
+		err := updateModelsSheet(srv, cred, sheetId, entriesToBeAddedInModelSheet)
+		if err != nil {
+			Log.Error(err)
```
lekaf974, line 151: Just to understand, we continue even if the update fail right ?

registry/spreadsheet_update.go
```diff
@@ -0,0 +1,349 @@
...
+
+		}
+	}
+
+	if modelCSVFilePath == "" || componentCSVFilePath == "" {
+		return "", "", "", ErrCSVFileNotFound(csvDirectory)
+	}
+	return modelCSVFilePath, componentCSVFilePath, relationshipCSVFilePath, nil
+}
+func getCSVHeader(filePath string) (headers, secondRow []string, err error) {
```
lekaf974, line 328: The function name is misleading since it does not return only the csv header

registry/spreadsheet_update.go
```diff
@@ -0,0 +1,349 @@
...
+	files, err := os.ReadDir(csvDirectory)
+	if err != nil {
+		return "", "", "", utils.ErrReadDir(err, csvDirectory)
+	}
+	var modelCSVFilePath, componentCSVFilePath, relationshipCSVFilePath string
+
+	for _, file := range files {
+		filePath := filepath.Join(csvDirectory, file.Name())
+		if !file.IsDir() && strings.HasSuffix(file.Name(), ".csv") {
+			headers, secondRow, err := getCSVHeader(filePath)
```
lekaf974, line 308: should we not evaluate error first ?

lekaf974 (changes requested):
Did a first pass and will continue soon

registry/helpers.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/layer5io/meshkit/logger"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/viper"
+)
+
+var (
+	// global logger variable
+	Log logger.Handler
+	//global logger error variable
+	LogError logger.Handler
```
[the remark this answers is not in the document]
riyaa14: https://github.com/meshery/meshery/blob/0ac709a34b0001b48bad6ed61dfe851fbb0036a8/mesheryctl/pkg/utils/helpers.go#L259 - that logger is under `utils` package in `mesheryctl` and this logger is associated with `registry` package in `meshkit`

registry/spreadsheet_update.go
riyaa14: yes, we log the error and continue with rest of the logic

riyaa14:
@lekaf974 Thank you for the review!
just to inform, I haven't made any changes to functions that I have "added" here
These files: model.go, relationship.go, component.go, spreadsheet_update.go are copied as it is from mesheryctl utils to here in meshkit.
Thus, I might not be sure of the reasoning behind certain logic written.

registry/spreadsheet_update.go
riyaa14: We are reading the [meshery integrations](https://docs.google.com/spreadsheets/d/1DZHnzxYWOlJ69Oguz4LkRVTFM79kC2tuvdwizOJmeMw/edit?usp=sharing) spreadsheet in which first two rows are headers, not just row one

That's why this function is named getCSVHeader, because row 1 and 2 are headers in this sheet
Screenshot 2025-02-26 at 2 29 17 AM

What do you think? shall we still name it something different?

registry/spreadsheet_update.go
riyaa14: yes, makes complete sense, moving the error up

registry/spreadsheet_update.go
riyaa14: According to my understanding, we can't because resetting the lists is based on conditions.
In the following condition only one list is updated - 
Screenshot 2025-02-26 at 2 43 37 AM

we aren't updating the lists after every iteration, it is conditional

registry/helpers.go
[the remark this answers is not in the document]
lekaf974: oh did not notice

registry/spreadsheet_update.go
lekaf974: Looking the spreadsheet, the real header look to be the second row only but if I am right test both in case someone use a copy were header is the first row right ?

riyaa14: Yes ✅️

registry/error.go
```diff
@@ -0,0 +1,79 @@
...
+	ErrMarshalStructToCSVCode     = "meshkit-11301"
+	ErrAppendToSheetCode          = "meshkit-11302"
+	ErrUpdateToSheetCode          = "meshkit-11303"
+	ErrFileReadCode               = "meshkit-11304"
+	ErrGeneratesModelCode         = "meshkit-11305"
+	ErrReadCSVRowCode             = "meshkit-11306"
+	ErrCSVFileNotFoundCode        = "meshkit-11307"
+	ErrUpdateComponentsCode       = "meshkit-11308"
+	ErrGeneratesComponentCode     = "meshkit-11309"
+	ErrUpdateRelationshipFileCode = "meshkit-11310"
```
lekaf974, line 19: Some of those error seems not to be specific to registry so wondering if they should not be moved to https://github.com/meshery/meshkit/blob/master/utils/error.go

lekaf974:
@riyaa14 took a look on other PR I am seeing that mesheryctl/internal/cli/root/model/generate.go was also changed due to this PR so changes related to models needs to be in models folder not i the registry one

riyaa14:
@lekaf974 the model generate shares one function with registry -> `getCSV`
model generate command first generates model, component and relationship CSVs from given spreadsheet. Then registers the models using these 3 CSVs (this is same as model import)

I can move this function to some common utility folder but since the function is tailored to meshery integrations sheet which is closely related to registry, getCSV should be a part of registry package in meshkit.

or we can move it to a common folder by naming it something like - `getCSVFromModelGenerationSheet`. What would you suggest?

registry/error.go
riyaa14: ErrMarshalStructToCSV, ErrFileRead, ErrReadCSVRow, ErrCSVFileNotFound -> these look like the ones that should be moved. right?

souvikinator: Please check if any of the errors already exists in the meshkit and use those instead

souvikinator (approved):
LGTM

riyaa14:
@souvikinator @lekaf974 Thanks for review, feedback incorporated ✅
Changes done - 

Errors related
- Errors related to CSVs moved to a new file created under `utils/csv` as `errors.go`
- ErrReadFile - moved to files package, as it seemed relevant there
-  `ErrCSVFileNotFoundCode` - this isn't moved as if you notice, its error messages mention model and component csvs, and thus specific to registry, thus shouldn't be moved to a common csv package

Correct placement for Contains function in registry/helpers.go
- Moved to utils package as it is a general function
- Changed name to Locate because Contains was already present, moreover, Locate seemed a better name for this function.
- Updated references in meshery and meshkit

files/error.go
```diff
@@ -9,6 +9,18 @@ import (
 	"github.com/layer5io/meshkit/errors"
 )
 
+var (
+	ErrFileReadCode = "meshkit-11304"
```
lekaf974, line 13: Can you move this with other variable declaration ?

files/error.go
```diff
@@ -9,6 +9,18 @@ import (
...
+)
+
+func ErrFileRead(err error) error {
```
lekaf974, line 16: Can you move this behind following variables declaration

utils/utils.go
```diff
@@ -261,6 +261,15 @@ func Contains[G []K, K comparable](slice G, ele K) bool {
 	return false
 }
 
+func Locate(key string, col []string) int {
```
lekaf974, line 264: Perhaps this could be renamed so it gives more details on the purpose o the function since it has become an utils function.

files/error.go
riyaa14: If you see the whole file, the variable declarations are separate for different error categories. 

Eg:

Here the error messages are specific to meshery designs
https://github.com/meshery/meshkit/blob/46ebe93256e39432588fe6e9c85cb19a909d2c5d/files/error.go#L24 

Here they are specific to archives
https://github.com/meshery/meshkit/blob/46ebe93256e39432588fe6e9c85cb19a909d2c5d/files/error.go#L330

and since ErrFileReadCode, is more generic and not related to these, I separated it out for better readability.

files/error.go
riyaa14: https://github.com/meshery/meshkit/pull/688#discussion_r1975867345

riyaa14:
@lekaf974 changes done as suggested ✅

lekaf974 (approved):
LGTM

PR #689, opened 2025-02-26
Enhance `Import Model` Modal

riyaa14:
**Description**

This PR fixes #

**Notes for Reviewers**
- More descriptive message for File Import 
- Disable URL as it isn't supported currently

## Before

https://github.com/user-attachments/assets/2901e317-8401-4c0a-a857-5935c2f02571

## After

https://github.com/user-attachments/assets/128130b6-3c1e-4d71-9baa-4ff974689628

**[Signed commits](../CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [x] Yes, I signed my commits.

PR #690, opened 2025-03-03
Revert "Enhance `Import Model` Modal"

riyaa14:
Reverts meshery/meshkit#689

Reverting this change because it was planned differently after discussion.

New changes - 
- URL import works fine, added detailed tooltip message for it
- schema moved and fetched from sistent

New PRs for this change - https://github.com/layer5io/sistent/pull/946 and https://github.com/meshery/meshery/pull/13806

### DIFF
--- a/schemas/configuration/modelImport.json
+++ b/schemas/configuration/modelImport.json
@@ -23,7 +23,7 @@
           "file": {
             "type": "string",
             "format": "file",
-            "description": "Browse the model file from your file system. Ensure it's an OCI artifact in .tar, .tar.gz, or .tgz format.",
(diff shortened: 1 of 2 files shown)


leecalcote:
@riyaa14 is this PR still relevant considering the centralization of UI schema in Sistent?

riyaa14:
@leecalcote if `meshery/meshkit` schemas aren't used anywhere else other than in `meshery/meshery` repo, then yes this PR isn't needed anymore. Just to be on the safe side, I reverted these schema changes since they might be used somewhere I am not aware of.

PR #707, opened 2025-04-02
changed the hardcoded format of spreadsheet API

### LINKED ISSUE #703: Issue in the hardcoded url format of Google spreadsheet API.
### Current Behavior

Currently we are hardcoding the url format like this
sheetURL = sheetURL + "/pub?output=csv" + "&gid=" + strconv.FormatInt(spreadsheetID, 10)
But this does not work as this is not meant for this but rather for publishing.
### Expected Behavior

The expected format is 
https://docs.google.com/spreadsheets/d/{spreadsheet-id}/gviz/tq?tqx=out:csv&gid=0
### Environment

- **Host OS:** Mac Linux Windows
- **Platform:** Docker or Kubernetes
- **Meshery Server Version:** stable-v
- **Meshery Client Version:** stable-v

---
### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://layer5.io/community/handbook)
- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/qJy1c95qirjgWQODApilR9)

VinayCheripally:
**Description**

This PR fixes #703 

**Notes for Reviewers**

### DIFF
--- a/registry/component.go
+++ b/registry/component.go
@@ -209,7 +209,7 @@ type ComponentCSVHelper struct {
 func NewComponentCSVHelper(sheetURL, spreadsheetName string, spreadsheetID int64, localCsvPath string) (*ComponentCSVHelper, error) {
 	var csvPath string
 	if localCsvPath == "" {
-		sheetURL = sheetURL + "/gviz/tq?tqx=out:csv" + "&gid=" + strconv.FormatInt(spreadsheetID, 10)
+		sheetURL = sheetURL + "/export?format=csv" + "&gid=" + strconv.FormatInt(spreadsheetID, 10)
 		Log.Info("Downloading CSV from: ", sheetURL)
 		dirPath := filepath.Join(utils.GetHome(), ".meshery", "content")
 		_ = os.MkdirAll(dirPath, 0755)
--- a/registry/component.go
+++ b/registry/component.go
@@ -216,7 +216,7 @@ func NewComponentCSVHelper(sheetURL, spreadsheetName string, spreadsheetID int64
 		csvPath = filepath.Join(dirPath, "components.csv")
 		err := utils.DownloadFile(csvPath, newSheetURL)
 		if err != nil {
-			// If downloading with the default URL fails, retry using the overridden URL.
+			// If downloading with the default URL Path fails, retry using the overridden URL Path.
 			newSheetURL = sheetURL + overridedURLPathAndQueryParams + strconv.FormatInt(spreadsheetID, 10)
 			err = utils.DownloadFile(csvPath, newSheetURL)
 			if err != nil {
--- a/registry/component.go
+++ b/registry/component.go
@@ -216,7 +216,8 @@ func NewComponentCSVHelper(sheetURL, spreadsheetName string, spreadsheetID int64
 		csvPath = filepath.Join(dirPath, "components.csv")
 		err := utils.DownloadFile(csvPath, newSheetURL)
 		if err != nil {
-			// If downloading with the default URL Path fails, retry using the overridden URL Path.
+			// The URL format /pub?output=csv (defaultURLPathAndQueryParams) fails for spreadsheets not published to the web.
+			// Instead, we use the /export?format=csv format (overridedURLPathAndQueryParams).
 			newSheetURL = sheetURL + overridedURLPathAndQueryParams + strconv.FormatInt(spreadsheetID, 10)
 			err = utils.DownloadFile(csvPath, newSheetURL)
 			if err != nil {
--- a/registry/component.go
+++ b/registry/component.go
@@ -216,6 +216,7 @@ func NewComponentCSVHelper(sheetURL, spreadsheetName string, spreadsheetID int64
 		csvPath = filepath.Join(dirPath, "components.csv")
 		err := utils.DownloadFile(csvPath, newSheetURL)
 		if err != nil {
+			// If downloading with the default URL fails, retry using the overridden URL.
 			newSheetURL = sheetURL + overridedURLPathAndQueryParams + strconv.FormatInt(spreadsheetID, 10)
 			err = utils.DownloadFile(csvPath, newSheetURL)
 			if err != nil {
--- a/registry/component.go
+++ b/registry/component.go
@@ -209,14 +209,14 @@ type ComponentCSVHelper struct {
 func NewComponentCSVHelper(sheetURL, spreadsheetName string, spreadsheetID int64, localCsvPath string) (*ComponentCSVHelper, error) {
 	var csvPath string
 	if localCsvPath == "" {
-		newSheetURL := sheetURL + "/pub?output=csv" + "&gid=" + strconv.FormatInt(spreadsheetID, 10)
+		newSheetURL := sheetURL + defaultURLPathAndQueryParams + strconv.FormatInt(spreadsheetID, 10)
 		Log.Info("Downloading CSV from: ", newSheetURL)
 		dirPath := filepath.Join(utils.GetHome(), ".meshery", "content")
 		_ = os.MkdirAll(dirPath, 0755)
 		csvPath = filepath.Join(dirPath, "components.csv")
 		err := utils.DownloadFile(csvPath, newSheetURL)
 		if err != nil {
-			newSheetURL = sheetURL + overrideURL + strconv.FormatInt(spreadsheetID, 10)
+			newSheetURL = sheetURL + overridedURLPathAndQueryParams + strconv.FormatInt(spreadsheetID, 10)
 			err = utils.DownloadFile(csvPath, newSheetURL)
 			if err != nil {
 				return nil, utils.ErrReadingRemoteFile(err)
--- a/registry/component.go
+++ b/registry/component.go
@@ -225,6 +225,7 @@ func NewComponentCSVHelper(sheetURL, spreadsheetName string, spreadsheetID int64
 				return nil, utils.ErrReadingRemoteFile(err)
 			}
 		}
+		sheetURL = newSheetURL
 	} else {
 		csvPath = localCsvPath
 	}
--- a/registry/component.go
+++ b/registry/component.go
@@ -209,23 +209,19 @@ type ComponentCSVHelper struct {
 func NewComponentCSVHelper(sheetURL, spreadsheetName string, spreadsheetID int64, localCsvPath string) (*ComponentCSVHelper, error) {
 	var csvPath string
 	if localCsvPath == "" {
-		newSheetURL := sheetURL + defaultURLPathAndQueryParams + strconv.FormatInt(spreadsheetID, 10)
-		Log.Info("Downloading CSV from: ", newSheetURL)
 		dirPath := filepath.Join(utils.GetHome(), ".meshery", "content")
-		_ = os.MkdirAll(dirPath, 0755)
-		csvPath = filepath.Join(dirPath, "components.csv")
-		err := utils.DownloadFile(csvPath, newSheetURL)
-		if err != nil {
-			// The URL format /pub?output=csv (defaultURLPathAndQueryParams) fails for spreadsheets not published to the web.
-			// mesheryctl registry publish and mesheryctl registry generate might fail If spreadsheets are not published.
-			// Instead, we use the /export?format=csv format (overridedURLPathAndQueryParams) which works for sheets that are shared with everyone.
-			newSheetURL = sheetURL + overridedURLPathAndQueryParams + strconv.FormatInt(spreadsheetID, 10)
-			err = utils.DownloadFile(csvPath, newSheetURL)
-			if err != nil {
-				return nil, utils.ErrReadingRemoteFile(err)
-			}
+		err := os.MkdirAll(dirPath, 0755)
+		if err != nil{
+			return nil, utils.ErrCreateDir(err, dirPath)
 		}
-		sheetURL = newSheetURL
+
+		csvPath = filepath.Join(dirPath, "components.csv")
+		
+		sheetURL,err = DownloadCSV(sheetURL,csvPath,spreadsheetID)
+		if err !=nil{
+			return nil,err
+		}
+
 	} else {
 		csvPath = localCsvPath
 	}
--- a/registry/component.go
+++ b/registry/component.go
@@ -209,7 +209,7 @@ type ComponentCSVHelper struct {
 func NewComponentCSVHelper(sheetURL, spreadsheetName string, spreadsheetID int64, localCsvPath string) (*ComponentCSVHelper, error) {
 	var csvPath string
 	if localCsvPath == "" {
-		sheetURL = sheetURL + "/pub?output=csv" + "&gid=" + strconv.FormatInt(spreadsheetID, 10)
+		sheetURL = sheetURL + "/gviz/tq?tqx=out:csv" + "&gid=" + strconv.FormatInt(spreadsheetID, 10)
 		Log.Info("Downloading CSV from: ", sheetURL)
 		dirPath := filepath.Join(utils.GetHome(), ".meshery", "content")
 		_ = os.MkdirAll(dirPath, 0755)
--- a/registry/component.go
+++ b/registry/component.go
@@ -217,7 +217,7 @@ func NewComponentCSVHelper(sheetURL, spreadsheetName string, spreadsheetID int64
 
 		csvPath = filepath.Join(dirPath, "components.csv")
 		
-		sheetURL,err = DownloadCSV(sheetURL,csvPath,spreadsheetID)
+		sheetURL,err = DownloadCSVAndGetDownloadURL(sheetURL,csvPath,spreadsheetID)
 		if err !=nil{
 			return nil,err
 		}
--- a/registry/component.go
+++ b/registry/component.go
@@ -209,7 +209,7 @@ type ComponentCSVHelper struct {
 func NewComponentCSVHelper(sheetURL, spreadsheetName string, spreadsheetID int64, localCsvPath string) (*ComponentCSVHelper, error) {
 	var csvPath string
 	if localCsvPath == "" {
-		sheetURL = sheetURL + "/export?format=csv" + "&gid=" + strconv.FormatInt(spreadsheetID, 10)
+		sheetURL = FormatSheetURL(sheetURL, spreadsheetID)
 		Log.Info("Downloading CSV from: ", sheetURL)
 		dirPath := filepath.Join(utils.GetHome(), ".meshery", "content")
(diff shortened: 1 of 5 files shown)


gourav-k-shaw:
LGTM.
@riyaa14 @lekaf974 you might have a feedback to offer

gourav-k-shaw (approved):

leecalcote:
Has this been tested with the current Meshery Integrations sheet?

lekaf974 (changes requested):
Added a comment with ideas to investigate

VinayCheripally:
@leecalcote I tested it and it is showing me that there zero models So  ,I'm trying to resolve that.

VinayCheripally:
@lekaf974 Yeah I think it makes sense to do what you are saying.I will work on them after I resolve this error with meshery Integration sheet.

VinayCheripally:
@leecalcote It seems the current URL format which is 
https://docs.google.com/spreadsheets/d/{spreadsheet-id}/gviz/tq?tqx=out:csv&gid={gid} is meant for visualization and is not meant for downloading the sheet as a csv file.
So I used the format https://docs.google.com/spreadsheets/d/{spreadsheet-id}/export?format=csv&gid={gid} which actually downloads the csv files.It works with both meshery Integrations sheet and also custom spreadsheets.
Screenshot from 2025-04-03 13-57-33
Screenshot from 2025-04-03 13-59-19

VinayCheripally:
@lekaf974 I will add a test file to this.But I have a doubt using the default one with custom spreadsheets gives an error directly is that fine?

lekaf974:
Can you share the error ?

riyaa14:
> Do we want to create a PR each time we will have to change it (if needed), we modify an hardcoded value by another harcoded one. What about to have a default value (which should be the current one) and add it override by an argument in the mehod ? This will allow to not break the current behavior and let the caller to override it on runtime if needed

I strongly agree with this point. Since there is no official documentation for this new URL format, it is good to have a default value and override it if needed.

leecalcote:
Is there are link pointing to where Google Sheets has changed their URL format?

VinayCheripally:
> Is there are link pointing to where Google Sheets has changed their URL format?

I did not find an official article for that change in format but I found this blog.
https://youneedawiki.com/blog/posts/google-sheets-url-parameters.html

riyaa14:
> Is there are link pointing to where Google Sheets has changed their URL format?

I couldn't find any documentation too regarding the change of URL format or even on what URL formats are valid. @VinayCheripally suggested some solutions which again aren't mentioned officially but rather from different sources on the net.

riyaa14:
> But I have a doubt ,using the default one with custom spreadsheets gives an error directly is that fine?

@VinayCheripally actually this was the original problem. The URL format was working for Meshery Integrations sheet which I suppose is because it was published a long time back, but for recently published sheets, the URL wasn't working

and this is a problem as:
1. Users won't be able to use meshery registry commands for bulk model generation
2. E2E tests would need a new sheet too, we can't use the Meshery Integrations sheet here.

riyaa14:
> This is the error we are getting right now when using custom spreadsheets.

@VinayCheripally what do you mean by custom spreadsheets here?

Also, if you see the url format you shared here, how is it again calling the original URL format : `/pub?output=csv&gid={sheet-id}`? Shouldn't this call the updated URL - `/export?format=csv&gid={sheet-id}` as per your changes in this PR.

lekaf974:
also could we validate there is no impact on other command before merging

VinayCheripally:
> > This is the error we are getting right now when using custom spreadsheets.
> 
> @VinayCheripally what do you mean by custom spreadsheets here?
> 
> Also, if you see the url format you shared here, how is it again calling the original URL format : `/pub?output=csv&gid={sheet-id}`? Shouldn't this call the updated URL - `/export?format=csv&gid={sheet-id}` as per your changes in this PR.

I was showing this to @lekaf974 because he asked what kind of error was coming if we use the /pub?output=csv as default and override it by /export?format=csv if we use an argument. So,If we use this without argument in our tests for some test-spreadsheet it would give us an error and I asked @lekaf974 if that was fine or not.

VinayCheripally:
> > This is the error we are getting right now when using custom spreadsheets.
> 
> @VinayCheripally what do you mean by custom spreadsheets here?
> 
> Also, if you see the url format you shared here, how is it again calling the original URL format : `/pub?output=csv&gid={sheet-id}`? Shouldn't this call the updated URL - `/export?format=csv&gid={sheet-id}` as per your changes in this PR.

By custom spreadsheets I meant something which user created by his own. Not the Meshery Integrations sheet.

VinayCheripally:
https://developers.google.com/workspace/drive/api/reference/rest/v3/files/export
This is the closest I could find there is a somewhat similar that is being used here.

VinayCheripally:
@lekaf974 @riyaa14 I have a doubt currently in meshkit for publishing the models to some path we are using the url 
https://docs.google.com/spreadsheets/d/[spreadsheet-id]/pub?output=csv&gid=[Sheet-ID]
I believe this is used when we publish a spreadsheet to web which does not need credentials to access as we have already published it.So why are we using CRED variable in the mesheryctl command.
Even for the url format 
https://docs.google.com/spreadsheets/d/{spreadsheet-id}/export?format=csv&gid=0
This url works only if the spreadsheet can be accessed by anyone with the link.It does not use any authorization.
The only way which uses a .json file is the one with google sheets API which is not used directly to download the CSV file but to just read the file.So ,I'm confused why we are using the base64 format of .json file.

lekaf974:
> @lekaf974 @riyaa14 I have a doubt currently in meshkit for publishing the models to some path we are using the url [https://docs.google.com/spreadsheets/d/[spreadsheet-id]/pub?output=csv&gid=[Sheet-ID]](https://docs.google.com/spreadsheets/d/%5Bspreadsheet-id%5D/pub?output=csv&gid=%5BSheet-ID%5D) I believe this is used when we publish a spreadsheet to web which does not need credentials to access as we have already published it.So why are we using CRED variable in the mesheryctl command. Even for the url format [https://docs.google.com/spreadsheets/d/{spreadsheet-id}/export?format=csv&gid=0](https://docs.google.com/spreadsheets/d/%7Bspreadsheet-id%7D/export?format=csv&gid=0) This url works only if the spreadsheet can be accessed by anyone with the link.It does not use any authorization. The only way which uses a .json file is the one with google sheets API which is not used directly to download the CSV file but to just read the file.So ,I'm confused why we are using the base64 format of .json file.

@VinayCheripally did not get involve in the command development but if you find some issues, create an issue with description then let's talk about it in slack and next dev meeting to get feedback.

I did not remind who worked on this implementation to give feedback, @leecalcote any idea ?

lekaf974:
Pretty sure doing those e2e tests implementation will help detect issues that we were not aware due to small test coverage

riyaa14:
> @lekaf974 @riyaa14 I have a doubt currently in meshkit for publishing the models to some path we are using the url [https://docs.google.com/spreadsheets/d/[spreadsheet-id]/pub?output=csv&gid=[Sheet-ID]](https://docs.google.com/spreadsheets/d/%5Bspreadsheet-id%5D/pub?output=csv&gid=%5BSheet-ID%5D) I believe this is used when we publish a spreadsheet to web which does not need credentials to access as we have already published it.So why are we using CRED variable in the mesheryctl command. Even for the url format [https://docs.google.com/spreadsheets/d/{spreadsheet-id}/export?format=csv&gid=0](https://docs.google.com/spreadsheets/d/%7Bspreadsheet-id%7D/export?format=csv&gid=0) This url works only if the spreadsheet can be accessed by anyone with the link.It does not use any authorization. The only way which uses a .json file is the one with google sheets API which is not used directly to download the CSV file but to just read the file.So ,I'm confused why we are using the base64 format of .json file.

@VinayCheripally spreadsheet-credentials (base64 encoded) are required for 
1. Getting the Sheet-IDs
2. In generate command, we update the components sheet and without credentials, we won't be able to edit the sheet

These are two areas where I found the use of CRED, might be more

VinayCheripally:
> https://developers.google.com/workspace/drive/api/reference/rest/v3/files/export This is the closest I could find about the URL format .there is a somewhat similar URL that is being explained here.

@riyaa14 Does this look similar?It has the same use case.can you look at it once?

VinayCheripally:
> @VinayCheripally
> 
> Some questions come to me
> 
>     1. We are making 3 times the same changes looks not a good practice IMO and not easy to maintain on long term
> 
>     2. Since it is in meshkit and it will be reused by external client (mesheryctl and server) to ensure it is not breaking them, I suggest to create a test
> 
>     3. Do we want to create a PR each time we will have to change it (if needed), we modify an hardcoded value by another harcoded one. What about to have a default value (which should be the current one) and add it override by an argument in the mehod ? This will allow to not break the current behavior and let the caller to override it on runtime if needed

I added a function in utils.go and used it at three places and by testing do you mean for only the function which I just changed or for all the functions in that model.go?

registry/utils.go
```diff
@@ -0,0 +1,7 @@
+package registry
+
+import "fmt"
+
+func FormatSheetURL(baseURL string, sheetID int64) string {
+	return fmt.Sprintf("%s/export?format=csv&gid=%d", baseURL, sheetID)
```
lekaf974, line 6: I don;t really get it, FormatSheetURL return the same hardcoded value from a function so what we want to achieve ?

VinayCheripally: As you said we were making the same changes three times , I thought using a function would make it easier to maintain?

riyaa14: @VinayCheripally we can define a variable in `registry/helper.go` for this url. and this variable can be used at all the places. A function doesn't seem like the right choice here.

lekaf974: +1 about last point.

You identified the issue, I was saying to keep functionallity as is and just allow url overriding where the current value doesn't work with what we already have. 

PS: Perhaps we can make it a little bit cleaner but it can be done in another issue.

registry/helpers.go
```diff
@@ -20,6 +20,8 @@ const (
 	artifactHub            = "artifacthub"
 	gitHub                 = "github"
 	shouldRegisterColIndex = -1
+	defaultURL             = "/pub?output=csv&gid="
+	overrideURL            = "/export?format=csv&gid="
```
lekaf974, line 24:
```suggestion
	defaultURLPathAndQueryParams             = "/pub?output=csv&gid="             = "/pub?output=csv&gid="
	overridedtURLPathAndQueryParams       = "/pub?output=csv&gid="            = "/export?format=csv&gid="
```

lekaf974 (changes requested):
Renaming proposal also how it will modify the behavior on the runtime when mesheryctl will trigger one of those method ?

VinayCheripally:
> Renaming proposal also how it will modify the behavior on the runtime when mesheryctl will trigger one of those method ?

So whenever the defaultURI returns an error from the downloadFile function it uses the overrideURL and calls the same function again and checks for errors and just returns if it works as intended or else returns the error.

VinayCheripally:
> > > Renaming proposal also how it will modify the behavior on the runtime when mesheryctl will trigger one of those method ?
> > 
> > 
> > So whenever the defaultURI returns an error from the downloadFile function it uses the overrideURL and calls the same function again and checks for errors and just returns if it works as intended or else returns the error.
> 
> Ok, did not realized it on first read, in this case here it is important to add a comment this behavior explaining why we do this, please.

got it ,I'll do that.

VinayCheripally:
@lekaf974 I added the comments.

registry/relationship.go
```diff
@@ -47,14 +47,19 @@ type RelationshipCSV struct {
...
+		newSheetURL := sheetURL + defaultURLPathAndQueryParams + strconv.FormatInt(spreadsheetID, 10)
+		Log.Info("Downloading CSV from: ", newSheetURL)
 		dirPath := filepath.Join(utils.GetHome(), ".meshery", "content")
 		_ = os.MkdirAll(dirPath, 0755)
 		csvPath = filepath.Join(dirPath, "relationship.csv")
-		err := utils.DownloadFile(csvPath, sheetURL)
+		err := utils.DownloadFile(csvPath, newSheetURL)
 		if err != nil {
-			return nil, utils.ErrReadingRemoteFile(err)
+			// If downloading with the default URL Path fails, retry using the overridden URL Path.
```
lekaf974, line 57: The comment says what the code is doing so it is not use full, please add context on why we have to do this patch

lekaf974 (changes requested):

registry/relationship.go
lekaf974: I mean we did this fix because some command were broken in mesheryctl, etc...

VinayCheripally: There are two commands which are broken because of this (registry publish and registry generate). I thought rather than adding the commands adding the core reason that why they are broken would be better, but yeah, I will add the commands just in case.

VinayCheripally: Added the commands in comments!

registry/component.go
```diff
@@ -209,14 +209,21 @@ type ComponentCSVHelper struct {
...
 		dirPath := filepath.Join(utils.GetHome(), ".meshery", "content")
 		_ = os.MkdirAll(dirPath, 0755)
 		csvPath = filepath.Join(dirPath, "components.csv")
-		err := utils.DownloadFile(csvPath, sheetURL)
+		err := utils.DownloadFile(csvPath, newSheetURL)
 		if err != nil {
-			return nil, utils.ErrReadingRemoteFile(err)
+			// The URL format /pub?output=csv (defaultURLPathAndQueryParams) fails for spreadsheets not published to the web.
+			// mesheryctl registry publish and mesheryctl registry generate might fail If spreadsheets are not published.
+			// Instead, we use the /export?format=csv format (overridedURLPathAndQueryParams) which works for sheets that are shared with everyone.
```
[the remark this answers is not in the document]
lekaf974: @VinayCheripally I had in mind this logic was embedded in a reusable function ?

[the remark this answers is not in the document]
VinayCheripally: @lekaf974 It's not embedded in a reusable function .Should I create a function in helpers.go and use it in all these three files?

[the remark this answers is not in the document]
VinayCheripally: @riyaa14 A doubt. Using the format /pub is not a deprecation right because it's meant for publishing it's just a bug right.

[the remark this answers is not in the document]
VinayCheripally: ok... didn't know that. I will update the comments.

registry/helpers.go
```diff
@@ -52,3 +56,26 @@ func SetupLogger(name string, debugLevel bool, output io.Writer) logger.Handler
...
+	// Additionally, the `/export?format=csv` format has the advantage of not requiring the sheet to be explicitly
+	// published to the web, unlike the default format which requires the sheet to be published in CSV format.
+	
+	if err !=nil{
+		newSheetURL = sheetURL + overridedURLPathAndQueryParams + strconv.FormatInt(spreadsheetID, 10)
+		err = utils.DownloadFile(csvPath, newSheetURL)
+		if err !=nil{
+			return "",utils.ErrReadingRemoteFile(err)
+		}
+	}
```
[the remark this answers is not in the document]
lekaf974: +1

registry/helpers.go
```diff
@@ -52,3 +56,26 @@ func SetupLogger(name string, debugLevel bool, output io.Writer) logger.Handler
 	}
 	return logger
 }
+
+// Downloads CSV file using spreadsheet URL
+func DownloadCSV(sheetURL ,csvPath string, spreadsheetID int64)  (string ,error){
+	newSheetURL := sheetURL + defaultURLPathAndQueryParams + strconv.FormatInt(spreadsheetID, 10)
+	Log.Info("Downloading CSV from: ", newSheetURL)
```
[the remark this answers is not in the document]
VinayCheripally: Yup that makes sense.

registry/relationship.go
```diff
@@ -47,14 +47,18 @@ type RelationshipCSV struct {
 func NewRelationshipCSVHelper(sheetURL, spreadsheetName string, spreadsheetID int64, localCsvPath string) (*RelationshipCSVHelper, error) {
 	var csvPath string
 	if localCsvPath == "" {
-		sheetURL = sheetURL + "/pub?output=csv" + "&gid=" + strconv.FormatInt(spreadsheetID, 10)
-		Log.Info("Downloading CSV from: ", sheetURL)
+		newSheetURL := sheetURL + defaultURLPathAndQueryParams + strconv.FormatInt(spreadsheetID, 10)
+		Log.Info("Downloading CSV from: ", newSheetURL)
```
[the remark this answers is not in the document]
VinayCheripally: Yeah removing it.

registry/helpers.go
```diff
@@ -52,3 +56,26 @@ func SetupLogger(name string, debugLevel bool, output io.Writer) logger.Handler
 	}
 	return logger
 }
+
+// Downloads CSV file using spreadsheet URL
+func DownloadCSV(sheetURL ,csvPath string, spreadsheetID int64)  (string ,error){
```
lekaf974, line 61: Just seen this why a Method called DownloadCSV is returning newSheetUrl ? the name is misleading

lekaf974: Also wondering if we are not downloading twice, one time to retrieve the sheet url in the url and another time when calling in each files

VinayCheripally: @riyaa14 @lekaf974 newSheetURL is being assigned to the variable sheetURL which is then used as spreadsheetURL while returning CSV helpers.
https://github.com/meshery/meshkit/blob/0bf035aa3b2328b0b8fbf217458a0dbf9f74a144/registry/model.go#L236

VinayCheripally: @lekaf974 Are you talking about how we are calling the functions from meshery repo ?

lekaf974: So like @riyaa14 mentioned remove the string returned since it is not used anymore

VinayCheripally: > @riyaa14 @lekaf974 newSheetURL is being assigned to the variable sheetURL which is then used as spreadsheetURL while returning CSV helpers.
> 
> https://github.com/meshery/meshkit/blob/0bf035aa3b2328b0b8fbf217458a0dbf9f74a144/registry/model.go#L236

@lekaf974 Here I already mentioned that it is being used as spreadsheetURL in initialized ModelCSVHelper.

lekaf974 (approved):

lekaf974:
@riyaa14 last check ok to merge then

PR #709, opened 2025-04-03
Improve error logs related to invalid source url in model generation

riyaa14:
**Description**

This PR fixes #

### Before

Screenshot 2025-04-03 at 9 05 15 PM

### After

Screenshot 2025-04-03 at 8 55 30 PM

**Notes for Reviewers**

- [*] Yes, I signed my commits.

### DIFF
--- a/generators/github/error.go
+++ b/generators/github/error.go
@@ -12,9 +12,9 @@ const (
 )
 
 func ErrGenerateGitHubPackage(err error, pkgName string) error {
-	return errors.New(ErrGenerateGitHubPackageCode, errors.Alert, []string{fmt.Sprintf("error generate package for %s", pkgName)}, []string{err.Error()}, []string{"invalid sourceurl provided", "repository might be private"}, []string{"provided sourceURL according to the format", "provide approparite credentials to clone a private repository"})
+	return errors.New(ErrGenerateGitHubPackageCode, errors.Alert, []string{fmt.Sprintf("Error generating package for %s", pkgName)}, []string{fmt.Sprintf("Error generating package for %s. %s", pkgName, err.Error())}, []string{"Invalid source URL provided", "Repository might be private"}, []string{"Ensure source URL follows the format: git://[owner]/[repositoryname]/[branch]/[path from the root of repository]", "Make sure the path to CRDs is correct in the URL", "Private repositories aren't supported currently, ensure that provided URL is from a public repository"})
 }
 
 func ErrInvalidGitHubSourceURL(err error) error {
-	return errors.New(ErrInvalidGitHubSourceURLCode, errors.Alert, []string{}, []string{err.Error()}, []string{"sourceURL provided might be invalid", "provided repo/version tag does not exist"}, []string{"ensure source url follows the format: git://<owner>/<repositoryname>/<branch>/<version>/<path from the root of repository>"})
+	return errors.New(ErrInvalidGitHubSourceURLCode, errors.Alert, []string{"Error parsing the source URL"}, []string{fmt.Sprintf("Error parsing the source URL: %s", err.Error())}, []string{"Source URL provided might be invalid", "Provided repo/version tag does not exist"}, []string{"Ensure source url follows the format: git://[owner]/[repositoryname]/[branch]/[path from the root of repository]"})
 }


leecalcote:
@riyaa14, thank you. This is an improvement. 🙂  From the looks of it, it seems that we can be more accurate with the suggested probable cause. At the point in time in our logic where we are unable to retrieve the latest release tag, it should be the case that we've already been able to reach the URL that the user has provided and have moved along in our logic to looking up the release tag. If this is true, suggesting that the #1 most probable cause is an invalid source URL is inaccurate, and not helpful to users resolving this issue, then retrying and being successful.

Here is the URL that I used when I received this error: git://github.com/Azure/azure-service-operator/blob/main/charts/azure-service-operator-1.9.0.tgz. Be a user. Try using this URL to create a model. When you get a failure message. What are you to do with the information provided in the Suggested Remediation?

As a user, can I point to a tarred Helm chart? This information isn't provided by the event. The event also does not link users to additional documentation on this topic, but should.

As a user, can I include "blob" or branch name ("main" or "master" or other) in the URL? I should be able to do so. We should not be placing this burden on the user. We can parse the URL intelligently.

@FaheemOnHub, where is the updated contributing docs on custom event notification formats? All events should include a link to a docs page. // @zihanKuang 

There are many other questions here that need answers. This discussion is still on the topic of our very first assignment: documentation of the various scenarios in which `mesheryctl registry` and `mesheryctl model` commands are run, and how their UX should improve. The flow for these scenarios is simply invoked from the UI, but the source code is the same centrally sourced code from MeshKit. 

The list of scenarios you've fathomed are now documented in the Meshery Test Plan spreadsheet as test cases, I think (awesome!). Please, please put your user hat on and attempt to use the Import Model and Create Model wizards - - with no assumption of prior knowledge. How can we improve? What doesn't make sense? 

// @ianrwhitney @lekaf974 @amitamrutiya

leecalcote:
Merge conflict

FaheemOnHub:
> @riyaa14, thank you. This is an improvement. 🙂 From the looks of it, it seems that we can be more accurate with the suggested probable cause. At the point in time in our logic where we are unable to retrieve the latest release tag, it should be the case that we've already been able to reach the URL that the user has provided and have moved along in our logic to looking up the release tag. If this is true, suggesting that the #1 most probable cause is an invalid source URL is inaccurate, and not helpful to users resolving this issue, then retrying and being successful.
> 
> Here is the URL that I used when I received this error: git://github.com/Azure/azure-service-operator/blob/main/charts/azure-service-operator-1.9.0.tgz. Be a user. Try using this URL to create a model. When you get a failure message. What are you to do with the information provided in the Suggested Remediation?
> 
> * As a user, can I point to a tarred Helm chart? This information isn't provided by the event. The event also does not link users to additional documentation on this topic, but should.
> * As a user, can I include "blob" or branch name ("main" or "master" or other) in the URL? I should be able to do so. We should not be placing this burden on the user. We can parse the URL intelligently with ease.
> * As a user, ....  _there are many other questions here that need answers. The two above are examples._
> 
> @riyaa14, this discussion is still on the topic of our very first assignment: documentation of the various scenarios in which `mesheryctl registry` and `mesheryctl model` commands are run, and how their UX should improve. The flow for these scenarios is simply invoked from the UI, but the source code is the same centrally sourced code from MeshKit.
> 
> Let's be discerning with our event notifications, ensuring that we are offering information back to the user at each step and specific to each step of the generation process.
> 
> @FaheemOnHub, where is the updated contributing docs on custom event notification formats? All events should include a link to a docs page, @riyaa14 @lekaf974 @alphaX86 @hexxdump. // @zihanKuang
> 
> The list of scenarios you've fathomed are now documented in the [Meshery Test Plan spreadsheet](https://docs.google.com/spreadsheets/d/13Ir4gfaKoAX9r8qYjAFFl_U9ntke4X5ndREY1T7bnVs/edit?gid=838298230#gid=838298230) as test cases, I think (awesome!). Please, please put your user hat on and attempt to use the Import Model and Create Model wizards - - with no assumption of prior knowledge. How can we improve? What doesn't make sense?
> 
> These need reconciled and improved:
> 
> 1. https://docs.meshery.io/guides/configuration-management/generating-models - missing the new import and create wizards found in the UI.
> 2. https://docs.meshery.io/guides/configuration-management/importing-models - missing an explanation from the perspective of the UI.
> 3. https://docs.meshery.io/project/contributing/contributing-models-quick-start - arguably the best, consolidated user guide, but is categorized under "Contributing Guides"
> 
> // @ianrwhitney @amitamrutiya

We have not made any new custom formatters , so I guess these docs are still in a good position : https://docs.meshery.io/project/contributing/contributing-ui-notification-center

PR #722, opened 2025-04-15
[model] Allow Random Model Names for Artifact Hub Registrant

Tanishq1604:
**Description**
This PR fixes an issue [#14282](https://github.com/meshery/meshery/issues/14282) where ArtifactHub registrant required model names to match package names exactly, while GitHub allowed arbitrary model names. The solution extracts package names from ArtifactHub URLs when possible, making the experience consistent between registrants.

## Implementation
Modified the ArtifactHub package manager to:
1. Parse query parameters from ArtifactHub URLs
2. Extract the `ts_query_web` parameter which contains the actual package name
3. Use this extracted name when searching for packages
4. Fall back to the original model name if no package name can be extracted

https://github.com/user-attachments/assets/0e01fa86-ae8d-4844-889a-f9f0f0437b5a

## Notes for Reviewers

This change makes ArtifactHub work like GitHub when it comes to model registration. Users can now provide arbitrary model names when using ArtifactHub registrant rather than requiring the model name to exactly match the ArtifactHub package name.


amitamrutiya:
@Tanishq1604 are we doing the same things for github? if yes then use the same handler for both. if no then we should do what we are doing for GitHub registrant.

leecalcote:
Lint failures...

hortison:
No activity here... PR abandoned?

Tanishq1604:
@amitamrutiya for now i have done for the artifact hub i am taking the name from the artifact hub url's query 
to make it like the github one would raises concern just like what @riyaa14 said.

amitamrutiya:
@Jougan-0 are we good to go here?

hortison:
PR abandoned by author.

Issue #755, opened 2025-07-09
Golangci-Lint Warnings and Deprecated API Usage


ShigrafS:
### ✅ **Current Behavior**

Running `make check` on the `meshkit` project produces a large number of warnings and suggestions from `golangci-lint`, including:

* **30 unchecked error return values** (e.g., `defer file.Close()`, `os.RemoveAll()`)
* **Use of deprecated packages or APIs**, such as:

  * `nats.EncodedConn` (no longer supported)
  * `opa/rego`, `opa/storage/inmem` (deprecated since OPA v1.x)
* **Improper formatting in error messages**, such as:

  * Capitalized error strings (`ST1005`)
  * Errors ending with punctuation
* **Ineffectual assignments** (e.g., assigned value to `err` never used)
* **Incorrect `Printf`-style formatting with `logrus.Info`**
* **Multiple redundant or outdated imports** (e.g., importing `"k8s.io/api/core/v1"` twice)
* **Staticcheck issues**, including:

  * Use of deprecated constants (`oras.PackManifestVersion1_1_RC4`)
  * Suggestions to use `strings.ReplaceAll` instead of `strings.Replace`
  * Opportunities to simplify code using tagged switches or merged assignments

These warnings are spread across a wide number of files including:

* `files/`, `models/`, `utils/`, `generator/`, `broker/`, and `cmd/`

Execution took **9+ minutes**, showing long analysis time due to complexity and volume.

---

### ✅ **Expected Behavior**

* `make check` should complete without producing lint warnings or errors.
* All error return values should be properly handled.
* No deprecated APIs or packages should be used.
* Error strings should follow Go formatting guidelines:

  * Begin with lowercase
  * Do not end in punctuation or newlines
* Code should follow best practices suggested by `staticcheck`, `govet`, `ineffassign`, and `errcheck`.
* The codebase should be fully compliant with the configured `.golangci.yml` rules.

---

### Screenshots/Logs
```
PS D:\Programs\meshkit\meshkit> make check GOLANGCI_LINT_EXTRA_ARGS="--timeout 30m"
golangci-lint run -c .golangci.yml -v --timeout 30m ./...
level=info msg="golangci-lint has version 2.2.0 built with go1.24.4 from 1e7781b6 on 2025-06-28T20:40:39Z"
level=info msg="[config_reader] Used config file .golangci.yml"
level=info msg="[config_reader] Module name \"github.com/meshery/meshkit\""
[...]

PR #757, opened 2025-07-10
fix: make model name comparison case-insensitive using EqualFold

Malka123456:
🔧 What this PR does

This pull request fixes an issue related to case-sensitive model name comparison in registry/model.go.
Previously, the model name comparison logic was case-sensitive:

if modelName != "" && modelName != model.Model

This caused the CLI flag --model kubernetes (or similar) to fail silently or behave incorrectly if the casing didn’t exactly match the internal model name.

🔄 Now it's been updated to use:

if modelName != "" && !strings.EqualFold(modelName, model.Model)

Using strings.EqualFold() makes the comparison case-insensitive, fixing the issue and improving user experience for commands using --model.
🐛 Fixes

    This addresses the issue where passing model names like --model Kubernetes or --model kubernetes wouldn't work as expected due to case mismatch.

    Related to Meshery CLI bug discussed in meshery/meshery#15339

✅ Changes Summary

    Updated conditional logic in registry/model.go inside InvokeGenerationFromSheet() function.

    Replaced modelName != model.Model with !strings.EqualFold(modelName, model.Model)

    Ensured backward compatibility and no impact on other flags.


registry/model.go
```diff
@@ -839,7 +839,7 @@ func InvokeGenerationFromSheet(wg *sync.WaitGroup, path string, modelsheetID, co
 	// Iterate models from the spreadsheet
 	for _, model := range modelCSVHelper.Models {
 
-		if modelName != "" && modelName != model.Model {
+		if modelName != "" && ! strings.EqualFold(modelName, model.Model) {
```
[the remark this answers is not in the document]
Malka123456: This is my mistake ,I added extra space I will shortly fix this.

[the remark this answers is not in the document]
Malka123456: Ok , I will consider updating in these lines  shortly .

Malka123456:
@leecalcote ,

I’ve made the required changes in this PR and resolved the DCO issue and made those changes which gemini-code-assist suggested to me. I’m open to feedback and ready to make improvements. I’d really appreciate it if you could take a look and guide me further.

Thanks for the opportunity to learn and contribute! 🙏

h origin fix-model-equalfold
```diff
@@ -0,0 +1,19 @@
+[33mcommit fd5d12635a8ddd7b603e44f09fe12bd86cc751be[m[33m ([m[1;36mHEAD[m[33m -> [m[1;32mfix-model-equalfold[m[33m)[m
```
n2h9: Hey hey, hello :wave: :innocent: 

Thank you for you contribution :bowing_woman: 

What is this file, is it a giff diff? Is it supposed to be apart of a pr?

registry/component.go
```diff
@@ -244,7 +244,7 @@ func (mch *ComponentCSVHelper) GetColumns() ([]string, error) {
 	return csvReader.ExtractCols(rowIndex)
 }
 
-func (mch *ComponentCSVHelper) ParseComponentsSheet(modelName string) error {
+func (mch *ComponentCSVHelper) ParseComponentsSheet(modelNames []string) error {
```
n2h9: Looking that you are changing  string to []string, you may consider  variadic function approach here: https://pkg.go.dev/github.com/rajkumar-km/go-play/go-basics/06-functions/variadic 

so you could maybe specify  smth like 
```modelNames ...string```

and it will be compatible with previous version.  
Even if we forget to update on the client side, it will still work.

File: registry/model.go [hunk 4]
```diff
@@ -29,6 +29,8 @@ import (
 	_model "github.com/meshery/schemas/models/v1beta1/model"
 	"github.com/meshery/schemas/models/v1beta1/subcategory"
 	log "github.com/sirupsen/logrus"
+	//showing error below import 
+	//"github.com/spf13/viper/internal/features"
```
n2h9, line 33: should we just remove this, or why did we add this comment?

h origin fix-model-equalfold
Malka123456: Is this you nikita?

registry/model.go
Malka123456: Yes i was also thinking to just remove this line bcz there was no use of this package, but it was already imported earlier,it was not showing any error and when i made changes it started showing error therefore i commented it. And I'm looking for someone review and can tell where is the issue.

registry/component.go
Malka123456: Thanks! That's a great point about backward compatibility and to avoid boilerplate — really appreciate the suggestion!

leecalcote:
@Malka123456 be sure to sign your commits - https://github.com/meshery/meshkit/pull/757/checks?check_run_id=46012452894

Namanv0509:
Thank You @Malka123456  for your contribution!!
If this work is complete, this is a good item to add to the weekly [Meshery Development meeting agenda](https://docs.google.com/document/d/15IEwtTxPOkQXN2r_1LzGXuAv2mIUlmbdtmxpyxXF-KI/edit). You can add this item in the doc, attend, and present it. Meeting details at https://meet.layer5.io/meshery.

h origin fix-model-equalfold
lekaf974, line 1: Can we remove this file ?

Malka123456:
@lekaf974  
I have removed that file. 
Pls review my pr again, would appreciate any changes

leecalcote:
Running build checks...

Malka123456:
Hey @leecalcote ,
I’ve checked and it looks like the test failures aren’t caused by my changes. I only updated component.go, model.go, and registry.go, and the failing tests don’t seem related to these files. Let me know if there’s anything I should fix on my end!

lekaf974:
For current failing tests there is this issue that should tackle them #755

registry/model.go
```diff
@@ -274,9 +276,18 @@ func (mch *ModelCSVHelper) ParseModelsSheet(parseForDocs bool, modelName string)
 		select {
 
 		case data := <-ch:
-			if modelName != "" && data.Model != modelName {
+			if len(modelNames) > 0 {
```
lekaf974, line 279: this logic is repeated multiple times in component.go and model.go, could we have a small method to avoid duplcaition ?

Malka123456: Good point! I'm thinking of extracting this into a small helper method to keep things clean. I see a possible approaches:

1. Create a standalone function like shouldSkipModel(...) that handles the condition.

Let me know if you have a preference or suggest a better alternative!

lekaf974: Approach LGTM

leecalcote:
Yay, @Malka123456! 🎈 

@Malka123456, make sure that Meshery Docs have been update to reflect this change, please.

PR #839, opened 2025-11-26
fix: Skip meshery-integration-template model during registry generate

Ontiomacer:
This PR fixes meshery/meshery #14702.

In the Meshery Integrations Spreadsheet, the row
meshery-integration-template is an example/template model used for
creating new model entries. It should not be processed during
mesheryctl registry generate.

This PR adds a simple conditional check inside registry/model.go to
skip this model gracefully, preventing the generator from attempting to
register a template model and avoiding unnecessary errors in the logs.
--------------------------------------------------------------------------------------
- Added conditional to skip model == "meshery-integration-template"
- Prevents generator from processing a template-only entry
- Eliminates the "generator not implemented" error during registry generate
---------------------------------------------------------------------------------------
What I changed :
- Added a conditional in registry/model.go to skip the model named meshery-integration-template.
- Prevents the registry generator from processing the template/example row.
- Removes the unnecessary “generator not implemented” errors during mesheryctl registry generate.
Ran:
mesheryctl registry generate --directory ./mesheryctl/templates/template-csvs

### DIFF
--- a/registry/model.go
+++ b/registry/model.go
@@ -31,7 +31,7 @@ import (
 	"golang.org/x/sync/semaphore"
 	"google.golang.org/api/sheets/v4"
 )
-const mesheryIntegrationTemplateModel = "meshery-integration-template"
+
 var modelToCompGenerateTracker = store.NewGenericThreadSafeStore[compGenerateTracker]()
 
 type compGenerateTracker struct {


Ontiomacer:
@leecalcote This PR resolves meshery/meshery#14702.

It adds a constant for meshery-integration-template and ensures this
template-only model is skipped during registry generation. Verified that
normal models continue to generate correctly.

Ontiomacer:
Any changes or issues with it @leecalcote

lekaf974:
attempt to fix meshery/meshery#14702

registry/model.go
```diff
@@ -838,10 +837,14 @@ func InvokeGenerationFromSheet(wg *sync.WaitGroup, path string, modelsheetID, co
 	}()
 	// Iterate models from the spreadsheet
 	for _, model := range modelCSVHelper.Models {
-
+		if model.Model == mesheryIntegrationTemplateModel{
```
lekaf974: 👋 @Ontiomacer thanks,

Even if it does the job, the intention is not clear enough, I mean if I am not aware of the open issue I will have the question `why do we skip this model`, I'd suggest to use a slice with a name that describe the intention to make easier to understand the main purpose. Then, you can check if the model name is inside that slice. Does it make sense ?

leecalcote: There is a "PublishToRegistry" column that should be set to false for this row and just like any other row with a false value for this column, those columns should not be considered for generation.

This column is an ideal flag (built for this purpose).

Ontiomacer: Thanks @leecalcote and @lekaf974!
Acknowledged — the PublishToRegistry flag is indeed the intended control for excluding template rows.
Since the spreadsheet doesn’t mark this row yet, the skip here prevents generator errors until the flag is updated.

marblom007: Matching on this model name makes for brittle code. Can we move to only processing components that have a `true` value for `PublishToRegistry`, instead?

Ontiomacer: @marblom007 
Thanks for the suggestion!
I understand the concern about brittleness.
Since changes related to PublishToRegistry fall under Meshery maintainers' decisions, I’ll leave this to the core team to incorporate as appropriate.
@leecalcote @lekaf974

leecalcote: > Matching on this model name makes for brittle code. Can we move to only processing components that have a `true` value for `PublishToRegistry`, instead?

This is the way forward.

leecalcote:
@Ontiomacer you can investigate the model generation errors seen in the nightly generator workflow. That would be helpful.

leecalcote:
@Ontiomacer implement this: "Can we move to only processing components that have a true value for PublishToRegistry, instead?" or close this PR.

Ontiomacer:
Thanks, @leecalcote — understood.

I’ll update this PR to implement the correct approach:
• Only process models where PublishToRegistry == true
• Ensure the template row and any other non-publishable rows are naturally excluded

I’ll also take a look at the nightly generator workflow errors and verify that the updated logic resolves them.

Will push the updated commit shortly.

Ontiomacer:
Update: Implemented PublishToRegistry-based filtering
Thanks for the feedback, everyone — especially @marblom007, @lekaf974, and @leecalcote.
I've updated the PR to follow the recommended approach:
✔ What changed
Removed the previous model-name–based skip.
The generator now processes only models where PublishToRegistry == true, exactly as suggested.
When a specific model is passed using --model, publish checks are bypassed so users can always generate for that model.
Added an informational log message when a model is skipped due to PublishToRegistry != true.
✔ Result
This removes the brittle string match and aligns the generator with the intended spreadsheet-driven behavior.
Please let me know if you'd like any additional refinements!

Rajesh-Nagarajan-11:
Thank you for your contribution! 🎉

If this work is completed, this would be a great item to add to the weekly **Meshery Development Meeting** [agenda](https://docs.google.com/document/d/15IEwtTxPOkQXN2r_1LzGXuAv2mIUlmbdtmxpyxXF-KI/edit?tab=t.0).  
Please update the meeting document, attend the meeting, and present your progress.

**Meeting Details**
- **Date:** December 03
- **Time:** 8:30 PM IST
- **Join Link:** https://meet.layer5.io/meshery

registry/model.go
```diff
@@ -32,7 +31,7 @@ import (
 	"golang.org/x/sync/semaphore"
 	"google.golang.org/api/sheets/v4"
 )
-
+const mesheryIntegrationTemplateModel = "meshery-integration-template"
```
leecalcote: This is not needed and goes against the objective of being "anti-brittle".

Ontiomacer:
Thanks for the review, @leecalcote
I’ve removed the unused mesheryIntegrationTemplateModel constant as suggested.
The loop logic already correctly filters models based on PublishToRegistry and modelName, so no additional template-specific handling was needed. The generation behavior remains consistent and anti-brittle.
The PR has been updated accordingly. Let me know if you'd like this check to be handled differently or validated through a helper utility.

leecalcote (approved):

PR #878, opened 2025-12-22
fix: include rendered templates in helm manifest generation

### LINKED ISSUE #877: GetManifestsFromHelm only returns CRDs and ignores chart templates
### Current Behavior
The function GetManifestsFromHelm currently only processes files in the `crds/` folder. It ignores the `templates/` folder, which leads to "no components found" errors during model generation because the main resources (Deployments, Services) are missing.

### Expected Behavior
The function should also render the `templates/` directory to include the main Kubernetes resources, not just the CRDs.

### Screenshots/Logs
```
Error generating model: artifact-hub
 no components found for model
Error generating model: argocd
 provided Chart URL for the model argocd is empty.
Error generating model: aws-load-balancer-controller
 no components found for model
Error generating model: ace
 no components found for model
Error generating model: ambassador
 no components found for model
```

parthivsaikia:
**Description**
This PR updates `GetManifestsFromHelm` to correctly render Helm chart templates instead of only retrieving CRDs. Previously, the function ignored the `templates/` directory, which caused "no components found" errors during model generation because standard resources (Deployments, Services, etc.) were missing.

This PR fixes #877 

### DIFF
--- a/.github/agents/meshkit-code-contributor.md
+++ b/.github/agents/meshkit-code-contributor.md
@@ -1,0 +1,202 @@
+---
+name: MeshKit Code Contributor Agent
+description: Expert-level Go engineering agent specialized in contributing to MeshKit, Meshery’s shared Go utilities and error handling library
+tools: [execute, read, edit, search, agent, web, todo]
+---
+
+# MeshKit Code Contributor
+
+You are an expert-level software engineering agent specialized in contributing to **MeshKit**, the shared Go library used across Meshery for error handling, utilities, and supporting tooling. MeshKit underpins consistent error semantics, code generation utilities, and other cross-project helpers used by Meshery Server, adapters, and related components.
+
+## Core Identity
+
+**Mission**: Deliver production-ready, maintainable Go code contributions to the MeshKit project that adhere to Meshery’s community standards, design principles, and architectural patterns. Execute systematically following Meshery contribution guidelines and operate autonomously to complete tasks.
+
+**Scope**: Contribute exclusively to MeshKit’s Go code and tooling, including:
+- **Error handling framework** and error code conventions
+- **Shared utilities** used by Meshery components
+- **Error code utility (errorutil)** for generating and analyzing error codes
+- **Tests, linting, and build system configuration** specific to MeshKit
+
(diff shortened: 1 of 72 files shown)


parthivsaikia:
I have used helm client in most places but I will try to refactor it further if possible.

leecalcote:
Good fix for a real gap in Helm manifest generation, @parthivsaikia! Including rendered templates alongside CRDs ensures component discovery actually works for standard Helm charts. Some areas to address:

1. **Leverage the Helm client more**: As @leecalcote noted, this is a lot of custom rendering logic for something the Helm Go client already handles. Specifically, the `repairYaml` function (injecting `---` before `apiVersion:` and deduplicating separators) suggests the upstream rendering output isn't being consumed correctly. Investigate whether using `action.Install` with `DryRun: true` and `ClientOnly: true` would give you properly separated YAML documents without needing the repair step.

2. **`repairYaml` is fragile**: Blindly inserting `\n---\n` before every `\napiVersion:` will break if `apiVersion:` appears inside a string value, a comment, or a multi-line field. If Helm's engine output genuinely concatenates documents without separators, file a bug upstream rather than papering over it — or at minimum add a comment explaining why this workaround is needed.

3. **`removeNonCrdValues` changed semantics**: The old function filtered out empty/null strings. The new version unmarshals JSON and only keeps documents with `kind: CustomResourceDefinition`. This is a narrower filter — it now drops all non-CRD resources. Make sure this is intentional and that callers of `GetCrdsFromHelm` don't expect other resource types.

4. **Hardcoded release options**: The release name `"meshery-helm-release"`, namespace `"default"`, and empty values are hardcoded. For a library function, consider accepting these as parameters (or via an options struct) so callers can customize rendering behavior. At minimum, document why these specific values were chosen.

5. **Import ordering**: The import block mixes standard library and third-party imports without a blank-line separator, which doesn't follow Go convention. Group them as: stdlib, then blank line, then third-party.

6. **Filter Helm test manifests**: The Gemini review suggested also filtering files containing `/tests/` in the path — you've incorporated this, which is good.

Your response about trying to use the Helm client more is the right direction.

PR #883, opened 2026-01-02
[Generators] Fix ExtractZip and implement robust recursive CRD discovery

### LINKED ISSUE #881: [Feature] Auto-discover CRDs from repository root URL
### Current Behavior

Currently, the model generator requires users to point to a specific directory inside of a repo to locate CustomResourceDefinitions (CRDs). This makes the source location brittle; if the repo structure changes, the generator fails.

### Desired Behavior

We need to make the location we use as the source of CRDs more robust.

Users should be able to simply point to a repository URL (e.g., https://github.com/crossplane/crossplane), and the component generator should automatically find the CRDs from there, regardless of where they are located in the file structure.

### Implementation
Use the logic in [crdsdev/doc](https://github.com/crdsdev/doc) as inspiration for enhancing the discovery mechanism in MeshKit. The goal is to move away from static directory paths to a more robust discovery method.

muhammadolammi:
**Description**

This PR fixes #881
It enhances MeshKit's component discovery mechanism by moving from brittle, static directory paths to a robust, recursive discovery method. It also fixes a critical bug in the ExtractZip utility that previously caused files to be extracted into the current working directory (CWD) rather than the intended temporary path. 

Key Changes:

1. Bug Fix: Robust ZIP Extraction
Path Correction: Fixed a bug where os.Mkdir used file.Name instead of filePath, causing directories to be created in the CWD. It now correctly uses filepath.Join(path, file.Name).

Resource Management: Refactored the extraction loop to use an anonymous closure. This ensures that defer fd.Close() and defer openedFile.Close() execute per file inside the archive, preventing "too many open files" errors and file descriptor leaks.

macOS Metadata Filtering: Added logic to skip macOS resource fork files (._ files), which were previously causing YAML unmarshal errors during component generation.

2. Feature: Recursive & Robust CRD Discovery
Empty Root/Branch Support: Updated the GitHub package manager to handle URLs without specific root paths (e.g., git://github.com/org/repo/).

 Case 1: Branch with no root - We set the root to "/**" to enable recursive scan
 Case 2: No Branch No root - We get the default branch from GitHub and set the root to "/**"

Enhanced Testing: Added a comprehensive suite of test cases to package_test.go to verify:

**Notes for Reviewers** : The refactor of ExtractZip was essential because the previous implementation failed when encountering archives with an internal directory (as seen in the Archive.zip test failures). By ensuring we create the full path via an if file.FileInfo().IsDir() {
				// this will create the folders in the filepath(/temp ) instead of cwd, which was a bug
				return os.MkdirAll(filePath, file.Mode())
			},
 we now safely handle archives that don't explicitly define directory entries for every file.

### DIFF
--- a/generators/github/error.go
+++ b/generators/github/error.go
@@ -20,7 +20,6 @@ func ErrInvalidGitHubSourceURL(err error) error {
 	return errors.New(ErrInvalidGitHubSourceURLCode, errors.Alert, []string{}, []string{err.Error()}, []string{"sourceURL provided might be invalid", "provided repo/version tag does not exist"}, []string{"ensure source url follows the format: git://<owner>/<repositoryname>/<branch>/<version>/<path from the root of repository>"})
 }
 
-// ErrGetDefaultBranch is the error for getting the default branch
 func ErrGetDefaultBranch(err error, owner, repo string) error {
 	return errors.New(ErrGetGithubDefaultBranchCode, errors.Alert, []string{fmt.Sprintf("Error while getting default branch for %s/%s", owner, repo)}, []string{err.Error()}, []string{"The repository might not exist"}, []string{})
 }


leecalcote:
Thanks, @muhammadolammi 👍

leecalcote:
@parthivsaikia, @marblom007, @aabidsofi19, please 👀 when convenient.

CWAbhi:
@muhammadolammi Please check who is assigned to the issue before raising it. Let’s discuss it first. I have already fixed this

generators/github/git_repo.go
```diff
@@ -32,10 +33,11 @@ func (gr GitRepo) GetContent() (models.Package, error) {
 	}
 
 	versions, err := utils.GetLatestReleaseTagsSorted(owner, repo)
-	if err != nil {
-		return nil, ErrInvalidGitHubSourceURL(err)
+	var version string
+	if err == nil && len(versions) > 0 {
```
parthivsaikia, line 39: Please also consider the case when err != nil from GetLatestReleaseTagsSorted

generators/github/git_repo_branch.go
```diff
@@ -0,0 +1,35 @@
...
+		return "", err
+	}
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != 200 {
+		return "", fmt.Errorf("github api: %d", resp.StatusCode)
```
parthivsaikia: For consistency use a meshkit error instead of fmt.Errorf

generators/github/git_repo_branch.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+	var out struct {
+		DefaultBranch string `json:"default_branch"`
+	}
+	if err := json.NewDecoder(resp.Body).Decode(&out); err != nil {
+		return "", err
```
parthivsaikia: same here. use a meshkit error

generators/github/git_repo_branch.go
```diff
@@ -0,0 +1,35 @@
+package github
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+func GetDefaultBranchFromGitHub(owner, repo string) (string, error) {
```
parthivsaikia: This function is just a wrapper around the getDefaultBranchFromGithub function. Delete this one and export the other.

generators/github/git_repo_branch.go
```diff
@@ -0,0 +1,35 @@
...
+	return getDefaultBranchFromGitHub(owner, repo)
+}
+
+func getDefaultBranchFromGitHub(owner, repo string) (string, error) {
```
parthivsaikia: I think refactoring this to accept a http client and base url as arguments will be a good idea so that we can test easily using httptest without making API calls to github which are slow and prone to failure.

generators/github/git_repo_branch_test.go
```diff
@@ -0,0 +1,48 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+
+			branch, err := GetDefaultBranchFromGitHub(tt.owner, tt.repo)
+
+			if (err != nil) != tt.wantErr {
+				t.Fatalf("unexpected error state: got err=%v, wantErr=%v", err, tt.wantErr)
+			}
+
+			if branch != tt.wantBranch {
+				t.Fatalf("branch: got %q want %q", branch, tt.wantBranch)
```
parthivsaikia: Instead of making api calls directly to github use `net/http/httptest` to mock the github api responses

muhammadolammi:
> @muhammadolammi Please check who is assigned to the issue before raising it. Let’s discuss it first. I have already fixed this

Ohh sorry brother, how did you go about it?

My current pr fail the linting check which im planning on solving, what do you think?

CWAbhi:
@muhammadolammi I appreciate solving this problem . About the linting i will check about it and raise a PR to fix it you can then check and comment on it.

muhammadolammi:
> @muhammadolammi I appreciate solving this problem . About the linting i will check about it and raise a PR to fix it you can then check and comment on it.

No thats not my point, im thinking of just stopping the pr so you can work on yours.

Its my pr that currently have issues now not yours.
Well done working boss.

CWAbhi:
@muhammadolammi  okay sure bro!

utils/unarchive.go
```diff
@@ -72,49 +72,51 @@ func readData(name string) (buffer []byte, err error) {
...
-		fd, err := file.Open()
-		defer func() {
-			_ = fd.Close()
-		}()
+	for _, file := range zipReader.File {
+		// We use a closure here so that 'defer' works correctly for each file
+		err := func() error {
+			fd, err := file.Open()
+			if err != nil {
+				return err
```
parthivsaikia, line 85: Use a meshkit error here.

generators/github/git_repo.go
muhammadolammi: okay for this, this is how we check i n walker  if g.referenceName != "" {
	// 	cloneOptions.ReferenceName = g.referenceName
	// } so i think we can still expect empty versions on error. After checking the version getter GetLatestReleaseTagsSorted if len(releases) == 0 {
		return nil, ErrGettingLatestReleaseTag(errors.New("no release found in this repository"))
	} Its retruning error on empty version. kindly update me on the desired behavior. No version we return an error? so even if its /owner/repo a release tag must be present (i will assume this is the behavior and add this check

generators/github/error.go
```diff
@@ -18,3 +19,8 @@ func ErrGenerateGitHubPackage(err error, pkgName string) error {
 func ErrInvalidGitHubSourceURL(err error) error {
 	return errors.New(ErrInvalidGitHubSourceURLCode, errors.Alert, []string{}, []string{err.Error()}, []string{"sourceURL provided might be invalid", "provided repo/version tag does not exist"}, []string{"ensure source url follows the format: git://<owner>/<repositoryname>/<branch>/<version>/<path from the root of repository>"})
 }
+
+// ErrGetDefaultBranch is the error for getting the default branch
```
lekaf974:
```suggestion
```

generators/github/git_repo.go
```diff
@@ -32,10 +34,17 @@ func (gr GitRepo) GetContent() (models.Package, error) {
 	}
 
 	versions, err := utils.GetLatestReleaseTagsSorted(owner, repo)
+
+	var version string
+	if err == nil && len(versions) > 0 {
+
+		version = versions[len(versions)-1]
+	}
```
lekaf974, line 42: this part should be done after `if err != nil {...}`

lekaf974 (changes requested):
Please revert all formatting changes not related to this PR to keep focus on real changes

muhammadolammi:
> Please revert all formatting changes not related to this PR to keep focus on real changes

Yeah, but some of the formatting issues will make "make test" fail. Or I can close this pr, find a commit that's stable on masters apply the change, then make a new pr.

PR #885, opened 2026-01-03
Fix/utils extract zip

muhammadolammi:
**Description**

This fixes the bug in the ExtractZip function; it currently creates all subfolders of the zip inside cwd when using "filename". By using filepath, we make sure it's produced in /tm,p, which other functions calling ExtractZip (like the ProcessContent) expect and depend on

**Notes for Reviewers**
This PR depends on "fix syntax error in registry/model.go #884" Please merge #884 first.

### DIFF
--- a/utils/zip_test.go
+++ b/utils/zip_test.go
@@ -4,6 +4,7 @@ import (
 	"archive/zip"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


leecalcote:
@parthivsaikia, perhaps, you might offer review or @lekaf974 might have a comment or two.

utils/zip_test.go
```diff
@@ -0,0 +1,108 @@
...
+				"._metadata":      "should skip",
+				"realfile.txt":    "should keep",
+			},
+			wantErr: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tmpDir, _ := os.MkdirTemp("", "extract-test-*")
```
parthivsaikia: I think it will be better if we don't ignore these errors.

utils/zip_test.go
```diff
@@ -0,0 +1,108 @@
...
+				f.Write([]byte(content))
+			}
+			writer.Close()
+			zipFile.Close()
+
+			err := ExtractZip(tmpDir, zipFile.Name())
+
+			if (err != nil) != tt.wantErr {
+				t.Errorf("ExtractZip() error = %v, wantErr %v", err, tt.wantErr)
+			}
```
parthivsaikia, line 78: The specific error message "zipslip: illegal file path" defined in the struct is not checked here. Please consider checking it too.

utils/unarchive.go
```diff
@@ -72,49 +73,58 @@ func readData(name string) (buffer []byte, err error) {
...
-		fd, err := file.Open()
-		defer func() {
-			_ = fd.Close()
-		}()
-
-		if err != nil {
-			return ErrExtractZip(err, path)
-		}
+			if !strings.HasPrefix(targetPath, destDir+string(os.PathSeparator)) && targetPath != destDir {
+				return fmt.Errorf("zipslip: illegal file path: %s", file.Name)
```
lekaf974, line 92: This must return a meshkit error, ErrExtractZip looks a good candidate

utils/unarchive.go
```diff
@@ -72,49 +73,58 @@ func readData(name string) (buffer []byte, err error) {
...
+			if strings.HasPrefix(filepath.Base(targetPath), "._") || filepath.Base(targetPath) == "__MACOSX" {
+				return nil
+			}
 
-		if file.FileInfo().IsDir() {
-			err := os.Mkdir(file.Name, file.Mode())
+			fd, err := file.Open()
 			if err != nil {
-				return ErrExtractZip(err, path)
+				return err
```
lekaf974, line 102: this must be a meskit error as well

muhammadolammi:
> Hi @muhammadolammi, thank you for this PR.
> 
> Please ensure all errors returned are meshkit errors
> 
> Documentation: https://docs.meshery.io/project/contributing/contributing-error

The errors in this function are wrapped with ErrExtractZip. The naked errors are in the closure function, where we call a function in the for loop that returns a Go error. Then, we check if it's a non-nil error and return an ErrExtractZip error of that error. I think it's a good design

utils/zip_test.go
muhammadolammi: done boss

leecalcote:
Well done on this fix, @muhammadolammi! The rewrite of `ExtractZip` is a real improvement — proper Zip Slip protection, correct resource cleanup with closures, and macOS metadata filtering. A few things to polish:

1. **Use meshkit errors consistently inside the closure**: The maintainers noted that naked `fmt.Errorf` calls inside the closure should be meshkit errors. Your response about the closure wrapping pattern is reasonable — the inner function returns a plain error, and the outer `if err != nil` wraps it with `ErrExtractZip`. However, consider whether the `zipslip: illegal file path` error specifically should carry its own meshkit error code, since it represents a distinct security-relevant failure mode that operators might want to alert on.

2. **Test for `__MACOSX` as a directory**: Your skip logic checks `filepath.Base(targetPath) == "__MACOSX"`, which works for a file literally named `__MACOSX`, but in real macOS-generated zips, `__MACOSX` is typically a directory containing `._*` files. The current code will skip the directory entry itself, but individual files inside it (like `__MACOSX/._readme.txt`) won't be caught because `filepath.Base` would return `._readme.txt` (which IS caught by the `._` prefix check) — so the behavior is likely correct, but add a test case with a nested `__MACOSX/subdir/._file` path to verify.

3. **Consider a size limit**: The old code used a small buffer (`1<<4` = 16 bytes) which was inefficient, and you've switched to `io.Copy` which uses a default 32KB buffer — good. But for defense-in-depth, consider adding a `maxFileSize` check or using `io.LimitReader` to prevent zip bombs from exhausting disk space.

The test coverage is solid — the Zip Slip attack test and destination correctness test are exactly what this code needs.

simihablo:
Merge conflict...

PR #886, opened 2026-01-03
Improved CRD discovery to automatically scan repositories

### LINKED ISSUE #881: [Feature] Auto-discover CRDs from repository root URL

CWAbhi:
**Description**
This PR fixes #881 

**Notes for Reviewers**
This PR enhances the CRD discovery mechanism by allowing users to specify a repository URL as the source instead of a static directory path. The generator will automatically discover CRDs within the repository, regardless of their location in the file structure.

**Implementation Details**
Introduces a more robust CRD discovery strategy inspired by the logic used in crdsdev/doc.
Replaces static directory assumptions with recursive repository scanning and pattern-based CRD detection.
Improves resilience against repository structure changes without impacting existing functionality.

**Benefits**
Reduces configuration complexity for users.
Makes CRD discovery resilient to repository refactors.
Improves usability and long-term maintainability of the model generator.

**Example Usage**
`Source: https://github.com/crossplane/crossplane`
The generator automatically locates CRDs without requiring users to specify internal paths.

- [ ✅] Yes, I signed my commits.

### DIFF
--- a/.github/agents/meshkit-code-contributor.md
+++ b/.github/agents/meshkit-code-contributor.md
@@ -1,7 +1,7 @@
 ---
 name: MeshKit Code Contributor Agent
 description: Expert-level Go engineering agent specialized in contributing to MeshKit, Meshery’s shared Go utilities and error handling library
-tools: [execute, read, edit, search, agent, web, todo]
+tools: [agent, edit, execute, memory, read, search, todo, web]
 ---
 
 # MeshKit Code Contributor


CWAbhi:
@leecalcote can you please check the PR and review it

leecalcote:
This is a good item to add to the weekly [Meshery Dev meeting agenda](https://meet.meshery.io/dev-minutes). You can add this item in the doc, attend, and present it. Meeting details at https://meshery.io/calendar.

CWAbhi:
@leecalcote Sounds good . I’ll add this to the agenda and join the meeting to present it. Thanks for the direction!

CWAbhi:
@leecalcote i have fixed the version issue kindly review it.

generators/github/git_repo.go
```diff
@@ -78,31 +94,153 @@ func (gr GitRepo) GetContent() (models.Package, error) {
...
+		}
+		
+		// Check each document for CRD
+		hasCRD := false
+		for _, doc := range documents {
+			doc = strings.TrimSpace(doc)
+			if doc == "" {
+				continue
+			}
+			// Check for YAML format
```
leecalcote, line 197: Use libraries, not string parsing.

generators/github/git_repo.go
```diff
@@ -78,31 +94,153 @@ func (gr GitRepo) GetContent() (models.Package, error) {
...
+	if size < 2 {
+		err = ErrInvalidGitHubSourceURL(fmt.Errorf("Source URL %s is invalid, must specify at least owner and repo", gr.URL.String()))
+		return
+	}
+	
+	owner = parts[0]
+	repo = parts[1]
+	
+	// Remove .git suffix from repo name if present
+	repo = strings.TrimSuffix(repo, ".git")
```
leecalcote: Let's find a library that offers this functionality.

CWAbhi: sure i will look for a library that offers this and update it

generators/github/git_repo.go
CWAbhi: okay noted !

generators/github/git_repo.go
parthivsaikia: Hi @CWAbhi you can check this library : https://github.com/git-download-manager/git-url-parse

generators/github/git_repo.go
```diff
@@ -78,31 +94,153 @@ func (gr GitRepo) GetContent() (models.Package, error) {
...
+			if strings.Contains(doc, "kind: CustomResourceDefinition") {
+				hasCRD = true
+				break
+			}
+			// Check for JSON format
+			if strings.Contains(doc, "\"kind\":\"CustomResourceDefinition\"") ||
+				strings.Contains(doc, `"kind":"CustomResourceDefinition"`) {
+				hasCRD = true
+				break
+			}
```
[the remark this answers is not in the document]
parthivsaikia: Consider these suggestions @CWAbhi

generators/github/scheme_interface.go
```diff
@@ -27,3 +36,27 @@ func NewDownloaderForScheme(scheme string, url *url.URL, packageName string) Dow
 	}
 	return nil
 }
+
+// isGitHubURL checks if the URL is a GitHub repository URL
+// This enables automatic CRD discovery for standard GitHub URLs
+func isGitHubURL(scheme string, url *url.URL) bool {
```
parthivsaikia: Using a library instead of manual parsing would be more robust.

generators/github/git_repo.go
CWAbhi: okay i will check this out @parthivsaikia .

generators/github/scheme_interface.go
CWAbhi: sure i will change from manual parsing of this to a library

.golangci.yml
```diff
@@ -2,12 +2,3 @@ version: "2"
...
   timeout: 5m
-
-linters-settings:
-  staticcheck:
-    go: "1.25"
-    checks: ["all", "-ST1005"]
-
-issues:
-  exclude:
-    - "ST1005"
```
parthivsaikia, line 13: Please consider reverting these changes since they are not focused on the specific issue and also they have been taken care in another PR.

CWAbhi: sure i ll revert the latest commit

parthivsaikia:
@CWAbhi have you addressed the changes requested by gemini? And also I cannot see any usage of libraries as requested.

CWAbhi:
@parthivsaikia i have not addressed the changes requested by gemini . but i have used the libraries as requested i have used different library but ill check it once again today and address the changes requested by gemini too and commit the changes.

CWAbhi:
@parthivsaikia can you check it once i have fixed the gemini and uses git parser library.

CWAbhi:
@leecalcote @parthivsaikia any updates on this?

leecalcote:
Good initiative tackling automatic CRD discovery, @CWAbhi! The integration of `git-url-parse` for URL handling and the `crdAwareFileInterceptor` concept are both sound. Several areas need attention:

1. **Fragile CRD detection**: The string-based check `kind: CustomResourceDefinition` works for basic cases but will miss variations in whitespace or quoting. The JSON check has a duplicated condition (both sides of the `||` are identical). As suggested in review, use regex patterns like `kind:\s*CustomResourceDefinition` for YAML and `"kind"\s*:\s*"CustomResourceDefinition"` for JSON — or better yet, do a lightweight YAML unmarshal to a struct with just a `Kind` field.

2. **Multi-document YAML splitting is overly complex**: The nested splitting logic (try `\n---\n` first, then fall back to line-by-line parsing) has edge cases — e.g., `---` at the start of a file, or the `len(docs) > 1` guard silently dropping single-document files. Consider using Go's `yaml.Decoder` in a loop instead, which handles multi-document YAML correctly out of the box.

3. **Revert unrelated `.golangci.yml` changes**: The PR strips the linter config down to just `run: timeout: 5m`, removing all staticcheck settings. This is unrelated to CRD discovery and has been flagged for revert in review comments. Please restore the original linter config.

4. **`isGitHubURL` doesn't use `scheme` parameter**: The function signature accepts `scheme` but never uses it. Remove it and update the call site.

5. **Address the overlap with PR #887**: Both this PR and @muhammadolammi's #887 are solving similar problems (recursive CRD discovery, URL parsing). Coordinate to avoid duplicate/conflicting work — the maintainers may want to merge one approach and close the other.

Keep at it — the CRD-aware interceptor pattern is a good idea that just needs more robust implementation.

PR #905, opened 2026-01-09
fix(tracing): disable tracing when no collector is configured

### LINKED ISSUE #894: Allow the configuration of distributed tracing to be disabled
### Current Behavior
Currently, while running Meshery Server, I see the following debug logs:

```
14:20:07 traces export: exporter export timeout: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp 127.0.0.1:4317: connect: connection refused"
```

I have not defined a trace collector in my local deployment. 

### Desired Behavior
I would like to be able to disable tracing in deployments without a defined collector target. These logs are unnecessary noise, when no collector is present.

### Implementation

1. Both github.com/meshery/meshkit **AND** github.com/meshery/meshery will need to be enhanced with this configurable deployment option.
2. Support an environment variable for Meshery deployments.
3. Ensure that this deployment option is exposed in Meshery's Helm chart values.yaml, too.

### Acceptance Tests

- [ ] 1. Unless, trace collector target is configured, don't attempt to connect to collector.
- [ ] 2. The appropriate installation / deployment pages in Meshery Docs are updated.

---
### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)
- 🎨 Wireframes and designs for Meshery UI in [Figma](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI)

iamgoeldhruv:
**Description**

### Summary
This PR adds support for explicitly disabling OpenTelemetry tracing when no collector
endpoint is configured.

### Changes
- Added an `Enabled` flag to tracing configuration
- Short-circuits tracer initialization when tracing is disabled
- Prevents exporter connection attempts when no collector is present
- Added unit tests to cover disabled tracing behavior

### Why this is needed
Currently, Meshery components attempt to connect to a default OTLP endpoint
even when no collector is configured, resulting in noisy timeout logs.
This change ensures tracing is a no-op unless explicitly enabled.

### Related issue
Fixes: meshery/meshkit#894

###Related PR
This issue is further addressed in meshery/meshery#16908.

### DIFF
--- a/tracing/tracing.go
+++ b/tracing/tracing.go
@@ -25,6 +25,9 @@ type Config struct {
 	Endpoint string `yaml:"endpoint" json:"endpoint"`
 	// Insecure determines whether to use an insecure connection (no TLS)
 	Insecure bool `yaml:"insecure" json:"insecure"`
+	// Adding extra flag to check if tracing is enabled
+	Enabled *bool `yaml:"enabled" json:"enabled"`
+
 }
 
 func InitTracerFromYamlConfig(ctx context.Context, config string) (*sdktrace.TracerProvider, error) {
@@ -43,6 +46,10 @@ func InitTracerFromYamlConfig(ctx context.Context, config string) (*sdktrace.Tra
 // It sets up OTLP gRPC exporter, resource attributes, and W3C trace context propagation
 func InitTracer(ctx context.Context, cfg Config) (*sdktrace.TracerProvider, error) {
 	// Validate configuration
+	if cfg.Enabled != nil && !*cfg.Enabled {
+    return nil, nil
+    }
+
 	if cfg.ServiceName == "" {
 		return nil, fmt.Errorf("service name is required")
 	}
--- a/tracing/tracing_test.go
+++ b/tracing/tracing_test.go
@@ -232,3 +232,20 @@ func TestNewTransport(t *testing.T) {
 		t.Error("NewTransport(customTransport) returned nil")
 	}
 }
+func TestInitTracerDisabled(t *testing.T) {
+	ctx := context.Background()
+	enabled := false
+
+	cfg := Config{
+		Enabled: &enabled,
+	}
+
+	tp, err := InitTracer(ctx, cfg)
+	if err != nil {
+		t.Fatalf("expected no error when tracing is disabled, got %v", err)
+	}
+
+	if tp != nil {
+		t.Fatalf("expected nil tracer provider when tracing is disabled")
+	}
+}


leecalcote:
Good start on this, @iamgoeldhruv! The `Enabled` flag approach is reasonable, but the implementation needs to go further to fully solve the problem:

1. **Doesn't solve the core issue**: The PR title says "disable tracing when no collector is configured," but the current code only handles the case where `Enabled` is explicitly set to `false`. If a config omits both `Enabled` and `Endpoint`, the function still proceeds and errors on the empty endpoint check at line 56. The fix should also return `nil, nil` when `cfg.Endpoint == ""` (i.e., treat missing endpoint as implicitly disabled), which was the original motivation.

2. **Fix indentation**: The new lines inside `InitTracer` use spaces instead of tabs, breaking the file's consistent tab indentation. This will likely show up as a linting issue.

3. **Improve the comment**: `// Adding extra flag to check if tracing is enabled` is redundant — the field name `Enabled` is self-explanatory. A better comment would describe the behavior: `// Enabled can be used to explicitly disable tracing. When nil, tracing is enabled if Endpoint is set.`

4. **Add a test for the missing-endpoint case**: Add a test where `Enabled` is nil and `Endpoint` is empty, verifying that tracing gracefully becomes a no-op instead of returning an error. This is the actual user scenario from the issue.

The direction is right — just needs to be completed end-to-end.

uzairhameed:
> @leecalcote here's my feedback:
>
> **1. Core logic is incomplete (critical)**
> The PR title says "disable tracing when no collector is configured" — but the current check only handles the case where `Enabled` is explicitly set to `false`. If a user omits both `Enabled` and `Endpoint` (the typical local dev setup from the issue), the code still proceeds and errors at the empty endpoint check. That's the exact noisy log scenario the issue describes.
>
> The fix should treat a missing endpoint as implicitly disabled:
> ```go
> if (cfg.Enabled != nil && !*cfg.Enabled) || cfg.Endpoint == "" {
>     return nil, nil
> }
> ```
>
> **2. Indentation**
> The new lines inside `InitTracer` use spaces instead of tabs, inconsistent with the rest of the file. This will likely fail linting.
>
> **3. Comment**
> `// Adding extra flag to check if tracing is enabled` is redundant. A better comment describes the behavior:
> ```go
> // If Enabled is explicitly false, or no endpoint is configured, tracing is a no-op.
> ```
>
> **4. Missing test case**
> There's no test for the actual scenario from the issue: `Enabled` is nil and `Endpoint` is empty. Please add a test verifying that this results in a no-op rather than an error.

---
>If I'm missing anything let me know. I'm happy to dig deeper and discuss further!

PR #932, opened 2026-03-11
Add embedded Meshery schema validation

marblom007:
## Summary
- add a reusable `schema` package for validating Meshery documents against embedded schemas from `github.com/meshery/schemas`
- register built-in schema mappings for relationships and other core document types, with schema-version detection and typed decode helpers
- normalize embedded schema loading for Meshery's OpenAPI-based assets and add package tests covering validation success/failure paths

## Testing
- go test ./schema -count=1
- golangci-lint run -c .golangci.yml ./schema/...

## Notes
- Full `make test` and `make check` still hit existing unrelated failures in the repository (for example `generators/github`, `registry`, and pre-existing repo-wide lint findings outside `schema/`).

### DIFF
--- a/schema/compiler.go
+++ b/schema/compiler.go
@@ -209,21 +209,3 @@ func jsonPointer(path []string) string {
 func escapeJSONPointerToken(token string) string {
 	return strings.NewReplacer("~", "~0", "/", "~1").Replace(token)
 }
-
-func keywordFromLocation(location string) string {
-	if index := strings.Index(location, "#"); index >= 0 {
-		location = location[index+1:]
-	}
-
-	location = strings.TrimPrefix(location, "/")
-	location = strings.TrimSuffix(location, "/")
-	if location == "" {
-		return ""
-	}
-
-	if index := strings.LastIndex(location, "/"); index >= 0 {
-		location = location[index+1:]
-	}
-
-	return strings.NewReplacer("~1", "/", "~0", "~").Replace(location)
-}


hortison (commented):
Each of the initial comments from gemini have been addressed by @simihablo, I think.

simihablo:
> Each of the initial comments from gemini have been addressed by @simihablo, I think.

jes, I see a new comment above. i will mark the others resolved and ask for another review.

schema/compiler.go
```diff
@@ -0,0 +1,192 @@
...
+import (
+	stderrors "errors"
+	"fmt"
+	"io/fs"
+	"log"
+	"net/url"
+	"strings"
+
+	"github.com/dlclark/regexp2"
+	jsonschema "github.com/santhosh-tekuri/jsonschema/v6"
```
aabidsofi19: i think it would be better to use openapi compatible validator

leecalcote: @aabidsofi19 as in kin-openapi, I assume. Is that the one that you're suggesting?

schema/registry.go
```diff
@@ -0,0 +1,124 @@
...
+	v.mu.Lock()
+	defer v.mu.Unlock()
+
+	if registration.Ref.SchemaVersion != "" {
+		v.registrations[schemaVersionKey(registration.Ref.SchemaVersion)] = registration
+	}
+
+	if registration.Ref.Type != "" {
+		v.registrations[typeKey(registration.Ref.Type)] = registration
+	}
```
PragalvaXFREZ, line 96: This is acceptable for now because we do not currently have multiple schema versions for the same type but this silently overwrites the type-based entry. When a second schema version is registered for the same type, the type-key lookup will return whichever was registered last. Should this be documented as intentional 'latest wins' behavior, or should it error?

schema/registry.go
```diff
@@ -0,0 +1,124 @@
...
+			Ref: Ref{
+				SchemaVersion: relationshipSchemaVersion,
+				Type:          TypeRelationship,
+			},
+			Location: relationshipSchemaLocation,
+		},
+		{
+			Ref: Ref{
+				Type: TypeEnvironment,
+			},
```
PragalvaXFREZ: Unlike all other builtins, this has no SchemaVersion, so auto-detection via Validate(data) won't work. Worth a comment explaining why.

schema/compiler.go
```diff
@@ -0,0 +1,192 @@
...
+	if cached, ok := v.cache.Load(location); ok {
+		return cached.(*jsonschema.Schema), nil
+	}
+
+	compiled, err, _ := v.compiling.Do(location, func() (any, error) {
+		if cached, ok := v.cache.Load(location); ok {
+			return cached.(*jsonschema.Schema), nil
+		}
+
+		compiler := jsonschema.NewCompiler()
```
guan404ming: `compile()` creates a new `jsonschema.NewCompiler()` every time. Shared $ref sub-schemas get reloaded repeatedly. Consider reusing a single compiler instance.

schema/validator.go
```diff
@@ -0,0 +1,340 @@
...
+	var document any
+
+	if err := meshkitencoding.Unmarshal(data, &document); err != nil {
+		return nil, err
+	}
+
+	return normalizeDocument(document)
+}
+
+func normalizeDocument(value any) (any, error) {
```
guan404ming, line 380: This does marshal/unmarshal round-trip even when input is already raw bytes from decodeDocument. That's three encode/decode passes per validation. Can we skip this when coming from ValidateBytes?

schema/validator.go
```diff
@@ -0,0 +1,340 @@
...
+	}
+
+	if base.Type == "" {
+		base.Type = fallback.Type
+	}
+
+	return base
+}
+
+func schemaVersionMismatch(expected string, document any) (Violation, bool) {
```
guan404ming, line 292: This pre-check exits early with one violation, hiding other field errors. If the schema already constrains schemaVersion, this is redundant. We could consider letting JSON Schema handle it uniformly.

schema/validator.go
```diff
@@ -0,0 +1,340 @@
...
+	return Default().detectRef(data)
+}
+
+func (v *Validator) detectRef(data []byte) (Ref, error) {
+	var header struct {
+		SchemaVersion string `json:"schemaVersion" yaml:"schemaVersion"`
+	}
+
+	if err := meshkitencoding.Unmarshal(data, &header); err != nil {
+		return Ref{}, ErrDecodeDocument(err)
```
PragalvaXFREZ, line 119: This parses the full document just to read schemaVersion. In the Validate() path, the same bytes are parsed again at line 170. Should we consider using decodeDocument once + detectRefFromDocument instead.

schema/validator.go
```diff
@@ -0,0 +1,340 @@
...
+	if ref.IsZero() {
+		return v.Validate(data)
+	}
+
+	registration, err := v.resolve(ref)
+	if err != nil {
+		return err
+	}
+
+	document, err := decodeDocument(data)
```
PragalvaXFREZ, line 198: Second parse of data when entering via Validate(). Could be avoided if Validate() parsed once and called detectRefFromDocument + validateDocument directly.

schema/compiler.go
guan404ming: I think in that context the pkg in that PR is suitable. `kin-openapi` needs an OpenAPI document to parse, but Meshery's schema files are plain JSON Schema Draft-07, not OpenAPI documents. kin-openapi's loader would fail on them directly.

schema/compiler.go
```diff
@@ -0,0 +1,179 @@
...
+)
+
+const embeddedSchemaScheme = "meshkit"
+
+type dlclarkRegexp regexp2.Regexp
+
+func (re *dlclarkRegexp) MatchString(value string) bool {
+	matched, err := (*regexp2.Regexp)(re).MatchString(value)
+	return err == nil && matched
+}
```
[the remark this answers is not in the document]
marblom007: Addressed by logging regexp2 match errors in schema/compiler.go.

schema/compiler.go
```diff
@@ -0,0 +1,179 @@
...
+func keywordFromLocation(location string) string {
+	location = strings.Trim(location, "#")
+	location = strings.Trim(location, "/")
+	if location == "" {
+		return ""
+	}
+
+	parts := strings.Split(location, "/")
+	return parts[len(parts)-1]
+}
```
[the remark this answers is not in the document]
marblom007: Addressed by RFC 6901-safe pointer handling and tests.

schema/validator.go
```diff
@@ -0,0 +1,323 @@
...
+		return err
+	}
+
+	document, err := normalizeDocument(value)
+	if err != nil {
+		return ErrDecodeDocument(err)
+	}
+
+	return v.validateDocument(registration, ref, document)
+}
```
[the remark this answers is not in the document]
marblom007: Addressed by normalizing once and detecting the ref from the normalized document.

schema/error.go
```diff
@@ -0,0 +1,120 @@
...
+		[]string{"Unable to compile schema"},
+		[]string{fmt.Sprintf("Failed to compile schema %q", location), err.Error()},
+		[]string{"The embedded schema or one of its references could not be resolved or contains unsupported syntax"},
+		[]string{"Ensure the schema location is registered correctly", "Verify that the referenced schema files are available in the embedded schemas module"},
+	)
+}
+
+func ErrValidateDocument(details ValidationDetails) error {
+	err := errValidateDocumentBase.ErrorV2(details)
+	return &err
```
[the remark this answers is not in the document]
marblom007: Addressed by teaching the errors helpers to read ErrorV2 values too.

schema/validator.go
```diff
@@ -0,0 +1,323 @@
...
+	v := &Validator{
+		fsys:          meshschemas.Schemas,
+		registrations: map[string]Registration{},
+	}
+
+	for _, registration := range builtinRegistrations() {
+		if err := v.Register(registration); err != nil {
+			panic(err)
+		}
+	}
```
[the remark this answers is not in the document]
marblom007: Addressed by making New return an error and adding MustNew for panic semantics.

schema/validator.go
```diff
@@ -0,0 +1,323 @@
...
+		return nil, err
+	}
+
+	var normalized any
+
+	if err := json.Unmarshal(normalizedBytes, &normalized); err != nil {
+		return nil, err
+	}
+
+	return normalized, nil
```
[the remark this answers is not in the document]
marblom007: Addressed by recursive normalization that preserves scalar types.

schema/compiler.go
```diff
@@ -0,0 +1,192 @@
...
+			SchemaPath:   schemaPath,
+			Keyword:      keywordFromLocation(schemaPath),
+			Message:      output.Error.String(),
+		})
+	}
+
+	for index := range output.Errors {
+		child := output.Errors[index]
+		collectViolations(&child, violations)
+	}
```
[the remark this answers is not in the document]
marblom007: Addressed; that traversal path was updated before the kin-openapi migration.

schema/compiler.go
marblom007: Addressed by switching the validator to kin-openapi.

schema/compiler.go
marblom007: Superseded by the kin-openapi migration; the jsonschema compiler path is gone.

schema/validator.go
marblom007: Addressed by the recursive normalizer, which avoids the old blanket round-trip.

schema/registry.go
suttonskate: Addressed in the latest update. I kept the current behavior but documented it explicitly in Register(): type-key registration is intentional "latest wins" for type-only lookups, and should be revisited if/when multiple schema versions per type are supported.

schema/registry.go
suttonskate: Added a comment on the builtin Environment registration explaining why it intentionally has no SchemaVersion today and therefore cannot be auto-detected via Validate(data); callers need to validate it explicitly by type.

schema/validator.go
suttonskate: Addressed by refactoring Validate() to decode once, then reuse the decoded document for detectRefFromDocument() and validateDocument(). detectRef(data) now also delegates through the shared decoded-document path.

schema/validator.go
suttonskate: Addressed together with the Validate() refactor: auto-detect + validate now reuse a single decoded document instead of parsing the input twice.

schema/compiler.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+func compileRegexp(pattern string) (openapi3.RegexMatcher, error) {
+	compiled, err := regexp2.Compile(pattern, regexp2.ECMAScript)
+	if err != nil {
+		return nil, err
+	}
+
+	return (*dlclarkRegexp)(compiled), nil
+}
```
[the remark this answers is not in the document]
suttonskate: Addressed by setting a default regexp2 MatchTimeout in compileRegexp() (currently 2s) and covering it with a unit test. That keeps regexp-based validation from running unbounded on pathological input.

schema/registry.go
```diff
@@ -0,0 +1,124 @@
...
+		if registration, ok := v.registrations[schemaVersionKey(ref.SchemaVersion)]; ok {
+			return registration, nil
+		}
+	}
+
+	if ref.Type != "" {
+		if registration, ok := v.registrations[typeKey(ref.Type)]; ok {
+			return registration, nil
+		}
+	}
```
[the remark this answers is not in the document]
suttonskate: Addressed by making SchemaVersion authoritative in resolve(): if SchemaVersion is explicitly provided but not registered, resolve() now returns ErrResolveSchema immediately instead of falling back to the type registration.

schema/validator.go
marblom007: I considered this one but left the pre-check in place for now. In the current schema set, schemaVersion is not enforced uniformly as a schema const/enum, so relying on JSON Schema alone would weaken the guard against validating a document against the wrong registered schema. The tradeoff is that this path can return the schemaVersion mismatch before unrelated field violations, but that still seemed preferable to silently accepting a mismatched version check depending on schema coverage.

errors/errors.go
simihablo, line 1: @copilot why are changes to this file necessary in context of this PR's purpose to add schema validation?

errors/errors.go
```diff
@@ -172,83 +173,79 @@ func (e *Error) ErrorV2(additionalInfo interface{}) ErrorV2 {
...
+	if stderrors.As(err, &errV1) && errV1 != nil {
+		return strings.Join(errV1.LongDescription, ".")
+	}
+
+	var errV2 *ErrorV2
+	if stderrors.As(err, &errV2) && errV2 != nil {
+		return strings.Join(errV2.LongDescription, ".")
 	}
 	return strings.Join(NoneString, "")
 }
```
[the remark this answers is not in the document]
marblom007: Addressed. I swapped the getter lookup order to prefer ErrorV2 over Error and added a regression test covering the case where an error can satisfy both lookups.

errors/errors.go
simihablo: @copilot do the changes in errors.go break compatibility with other Meshery components that are currently using this library for error reporting?

leecalcote:
Why can't this package import the various versioned schemas from the released meshery/schemas package, then  programmatically discern and align schemas to validators? This is completely possible. Do this. Don't create fragile, high maintenance code.

schema/registry.go
```diff
@@ -0,0 +1,137 @@
...
+				Type:          TypeRelationship,
+			},
+			Location: relationshipSchemaLocation,
+		},
+		{
+			// Environment schema has no SchemaVersion intentionally: the upstream
+			// schemas module does not yet define an EnvironmentSchemaVersion constant.
+			// As a result, auto-detection via Validate(data) is not supported for
+			// environment documents; callers must use ValidateBytes(Ref{Type: TypeEnvironment}, data)
+			// or ValidateAs(TypeEnvironment, data) explicitly.
```
leecalcote: This PR https://github.com/meshery/schemas/pull/621 adds schemaVersion to the Environment construct.

simihablo: I think that we should await a release of meshery/schemas, then.

errors/errors.go
suttonskate: Needed because schema validation returns ErrorV2 and these helpers feed logger metadata; the change is additive and covered by regression tests.

schema/registry.go
suttonskate: Addressed by dynamic discovery: once the released schemas package includes Environment schemaVersion, MeshKit will pick it up automatically without another code change.

suttonskate:
I upgraded to latest Meshkit v0.8.127 w/Environment schemaVersion

schema/registry.go
```diff
@@ -0,0 +1,527 @@
...
+
+func discoverRegistration(fsys fs.FS, assetPath string) (Registration, bool, error) {
+	asset, err := fs.ReadFile(fsys, assetPath)
+	if err != nil {
+		return Registration{}, false, err
+	}
+
+	var document any
+	if err := meshkitencoding.Unmarshal(asset, &document); err != nil {
+		return Registration{}, false, nil
```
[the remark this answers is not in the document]
simihablo: @copilot, read the coding agent in this repo. We do NOT use `fmt` for errors, but use MeshKit's error functions.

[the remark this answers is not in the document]
suttonskate: Addressed in ec5ca61: schema asset unmarshal failures now return a MeshKit schema error, and the remaining formatting stays inside MeshKit structured errors.

schema/compiler.go
```diff
@@ -0,0 +1,224 @@
...
+	// defaultRegexpMatchTimeout caps regexp2 backtracking evaluation to guard
+	// against potential CPU exhaustion when matching untrusted input against
+	// complex patterns. regexp2 uses a backtracking engine (unlike Go's RE2),
+	// so unbounded matching can become a denial-of-service vector.
+	defaultRegexpMatchTimeout = 2 * time.Second
+)
+
+type dlclarkRegexp regexp2.Regexp
+
+var regexpMatchStringErrorf = log.Printf
```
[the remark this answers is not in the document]
suttonskate: Addressed in c83d6ce: exported RegexpMatchStringErrorf so consumers can route these warnings through their own logger.

PR #943, opened 2026-03-16
Remove explicit status: enabled from generated component definitions

megha1807:
## Description

This PR removes the explicit inclusion of status: enabled from generated component definitions.

Since `enabled` is already the default behavior for components, explicitly adding `status: enabled` in the generated JSON definitions introduces unnecessary redundancy. This update modifies the component generation logic so that the `status` field is only included when the status differs from the default value.

With this change:

* status will not be included when the value is enabled.
* status will only be added when the value is ignored or disabled.

This reduces unnecessary properties in generated component definitions while preserving correct behavior.

Fixes #17918

## Notes for Reviewers

* Updated component generation logic in registry/component.go.
* Ensures that the `status` field is omitted when it matches the default (`enabled`).
* Keeps the `status` field for non-default values (`ignored`, `disabled`).
* Improves clarity and reduces redundancy in generated component definitions.

## Signed Commits
* ✅ Yes, I signed my commits.

### DIFF
--- a/registry/component.go
+++ b/registry/component.go
@@ -58,26 +58,34 @@ type ComponentCSV struct {
 // The Component Definition generated assumes or is only for components which have registrant as "meshery"
 func (c *ComponentCSV) CreateComponentDefinition(isModelPublished bool, defVersion string) (component.ComponentDefinition, error) {
 	status := entity.Enabled
-	if c.Status != "" {
-		if utils.ReplaceSpacesAndConvertToLowercase(c.Status) == "ignored" {
-			status = entity.Ignored
-		}
+
+if c.Status != "" {
+	switch utils.ReplaceSpacesAndConvertToLowercase(c.Status) {
+	case "ignored", "disabled":
+		status = entity.Ignored
+	case "enabled":
+		status = entity.Enabled
 	}
-	componentDefinition := &component.ComponentDefinition{
-		SchemaVersion: schmeaVersion.ComponentSchemaVersion,
-		DisplayName:   c.Component,
-		Format:        "JSON",
-		Version:       defVersion,
-		Metadata: component.ComponentDefinition_Metadata{
-			Published: isModelPublished,
-		},
-		Status: (*component.ComponentDefinitionStatus)(&status),
-		Component: component.Component{
-			Kind:    c.Component,
-			Schema:  c.Schema,
-			Version: c.Version,
-		},
-	}
+}
+
+componentDefinition := &component.ComponentDefinition{
+	SchemaVersion: schmeaVersion.ComponentSchemaVersion,
+	DisplayName:   c.Component,
+	Format:        "JSON",
+	Version:       defVersion,
+	Metadata: component.ComponentDefinition_Metadata{
+		Published: isModelPublished,
+	},
+	Component: component.Component{
+		Kind:    c.Component,
+		Schema:  c.Schema,
+		Version: c.Version,
+	},
+}
+
+if status != entity.Enabled {
+	componentDefinition.Status = (*component.ComponentDefinitionStatus)(&status)
+}
 	if c.Description != "" {
 		componentDefinition.Description = c.Description
 	}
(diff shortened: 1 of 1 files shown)


Bhumikagarggg:
@megha1807 Thank you for your contribution! Let's discuss this during the website call today at 5:30 PM IST | 6 AM CST Add it as an agenda item to the [meeting minutes](https://docs.google.com/document/d/1XczAHXVe2FIWPqiF57ospJ43zw5cZQ7ui8mn39v5EvA/edit?usp=drivesdk), if you would 🙂

leecalcote:
Good work! You might like to present your work in the upcoming, weekly Meshery development meeting ([meshery.io/calendar](meshery.io/calendar)). If so, I encourage you to add this item in the meeting agenda ([meet.meshery.io/dev-minutes](https://meet.meshery.io/dev-minutes)), attend, present, and receive feedback.

@megha1807, will you please share a model that has been generated with this code?

megha1807:
Thank you! @leecalcote  I’d be happy to present this in the upcoming Meshery development meeting—I’ll add it to the agenda.

Regarding the model, I’ll generate one using the current code and share it here shortly for your review.

megha1807:
Hello @aabidsofi19 
could you please let me know if any additional changes are needed for this PR, or if it’s ready to be merged? 
Thanks!

leecalcote:
> Hello @aabidsofi19 could you please let me know if any additional changes are needed for this PR, or if it’s ready to be merged? Thanks!

@megha1807, Yes: share a model that has been generated with this code.

megha1807:
@leecalcote  

As requested, I tested the updated `CreateComponentDefinition` logic locally by generating component definitions using a small test program invoking the updated function.

**Default case (`enabled`):**
```json
{
  "displayName": "sample-component"
}
```

**Non-default case (`disabled` → mapped to `ignored`):**
```json
{
  "displayName": "sample-component",
  "status": "ignored"
}
```

This confirms that:
- `status` is omitted when it is `enabled` (default)
- `status` is included only when it differs from the default value
- existing status mapping behavior is preserved

leecalcote:
Nice focused change, @megha1807! The intent is clear and the approach is right. A couple of things to tighten up:

1. **Indentation is broken**: The new `switch` block and `componentDefinition` struct literal lost their indentation — they're flush-left instead of being indented inside the function body. This needs to be fixed to match the rest of the file's style.

2. **Restore status handling in `UpdateCompDefinition`**: You removed the status update logic from `UpdateCompDefinition` entirely, but this means existing components can no longer have their status updated from the CSV (e.g., if a component changes from `ignored` back to `enabled`). As the Gemini review noted, consider reintroducing the logic there — and to avoid duplication, extract a small helper like `statusFromCSV()` that both functions share.

3. **Simplify the switch**: The `case "enabled"` arm is redundant since `status` is already initialized to `entity.Enabled`. You can also drop the outer `if c.Status != ""` guard — if it's empty, the switch simply won't match anything.

4. **Share a generated model output**: @leecalcote asked for an actual generated model file — you provided inline JSON snippets which is helpful, but sharing a complete generated `.json` file from a real model would give reviewers full confidence.

Almost there — just clean up the formatting and restore the update path.

Siddhi-sahu:
> @leecalcote
> 
> As requested, I tested the updated `CreateComponentDefinition` logic locally by generating component definitions using a small test program invoking the updated function.
> 
> **Default case (`enabled`):**
> 
> ```json
> {
>   "displayName": "sample-component"
> }
> ```
> 
> **Non-default case (`disabled` → mapped to `ignored`):**
> 
> ```json
> {
>   "displayName": "sample-component",
>   "status": "ignored"
> }
> ```
> 
> This confirms that:
> 
> * `status` is omitted when it is `enabled` (default)
> * `status` is included only when it differs from the default value
> * existing status mapping behavior is preserved

hello @megha1807, did you validate this change with the cli? please address @leecalcote's comments and share an actual generated model which was created without the status property. Thank you.

leecalcote:
@megha1807 still working on this?

PR #951, opened 2026-03-28
Remove redundant status: enabled from generated components

BasavarajBankolli:
**Description**

This PR fixes [17918](https://github.com/meshery/meshery/issues/17918)

Previously, generated component definitions included `"status": "enabled"` by default. Since `"enabled"` is already the implicit default behavior, explicitly including it results in redundant data.

This PR updates the model generation logic in MeshKit to:
- Avoid assigning `"status": "enabled"` by default
- Only include the `status` field when it is explicitly set to a non-default value (e.g., `"ignored"`)

**Notes for Reviewers**

- Updated logic in `registry/component.go` and `registry/model.go` to prevent default status injection.
- Ensured that `status` is only assigned when required.
- Verified that generated component definitions no longer include `"status": "enabled"`.
- No changes to behavior when status is explicitly set (e.g., `"ignored"` remains unaffected).


BasavarajBankolli:
Hi @Siddhi-sahu ,

I’ve validated the changes locally by linking my modified meshkit and regenerating models.
Key observations:

image

1. In model.json:
   - "status" is now empty ("") instead of defaulting to "enabled"

2. In component definition:
   - "status" is null, confirming that component status is no longer redundantly inherited from the model
   
validate

This confirms that redundant status propagation has been removed and defaults are handled correctly.
Please find the attached screenshots highlighting these changes.


Thank you

lekaf974:
Please add unit tests here

lekaf974:
what about the other status, why only  support `ignored` ?

leecalcote:
@BasavarajBankolli are you still working on this?

BasavarajBankolli:
Hi @leecalcote, yes, I’m actively working on this.

I’ve addressed the feedback regarding generic status handling (not limited to "ignored") and added unit tests to cover multiple cases including empty, enabled, ignored, and other statuses.

I’m currently validating the tests and will push the updated changes shortly.

Thank you for your patience!

PR #952, opened 2026-03-28
[Registry] Cache shared manifest fetches during generation

hssuryawanshib23-afk:
## Summary
Cache package fetches during registry generation when multiple models share the same registrant and source URL.

## Changes
- add a per-generation package fetcher keyed by registrant and source URL
- use singleflight so concurrent models sharing the same source only fetch once
- reuse the fetched package across model generation instead of calling `GetPackage()` for every model
- add regression tests for repeated, concurrent, and cross-registrant fetch behavior

## Why
ASO model generation currently fetches the same shared manifest once per model even though all models point to the same source URL. This change deduplicates those fetches without changing generated output.

## Verification
- `go test ./registry/... -count=1`

## Related
- meshery/meshery#18286
- part of meshery/meshery#11244


leecalcote:
### Code review

The caching approach using `singleflight` + `sync.Map` is well-structured, and the test coverage for the deduplication and concurrency behaviors is thorough.

Found 1 issue:

1. `modelName` is excluded from the cache key, but both generator backends use it to determine which package to fetch

The cache key is built from only `registrant` and `sourceURL`:

https://github.com/meshery/meshkit/blob/36ab6bbef69d0b9cbed9b62aa980bdd07437e3e4/registry/model.go#L54-L56

However, `modelName` is passed to `newGenerator` and both backends use it materially:

- **ArtifactHub** calls `GetAhPackagesWithName(ahpm.PackageName)` — it searches ArtifactHub by name, so different model names at the same source URL return different packages.
- **GitHub** sets `Name: packageName` on the returned `GitHubPackage`, which propagates into generated components.

The result: when multiple models share the same `(registrant, sourceURL)` but have different names, every model after the first silently receives the cached package that was fetched using the first model's name. For ArtifactHub registrants in particular, this means the wrong package data is used for generation.

Fix: include `modelName` (normalized, for consistency with `registrant`) in the cache key, e.g. `fmt.Sprintf("%s\x00%s\x00%s", registrant, sourceURL, modelName)`. This trades some cache efficiency for correctness, though in practice most sharing happens across models that differ only in name at the same source URL — meaning the cache would rarely be hit anyway.

hssuryawanshib23-afk:
Thanks, @leecalcote. Updated this to address the `modelName` correctness issue.

  I split the behavior by backend:
  - `ArtifactHub` now includes `modelName` in the cache key, since package resolution there is name-
  sensitive
  - `GitHub` still deduplicates by shared `registrant + sourceURL`, but returns a per-model copy with
  the requested name so generated component metadata stays correct

  So this keeps the intended optimization for shared GitHub manifests like ASO, while avoiding
  incorrect package reuse for cases where the model name materially affects package selection.

  I also updated the tests to cover both behaviors and re-verified locally with:
  - `go test ./registry/... -count=1`

leecalcote:
Thanks!

PR #986, opened 2026-04-22
fall back to kubeconfig loader behavior (exec plugin capable)

### LINKED ISSUE #985: In-cluster installation of Meshery on EKS fail for clusters that use exec-based kubeconfig auth
### Current Behavior
`mesheryctl system start -p kubernentes` errors out because `ApplyHelmChart` fails against EKS clusters that use exec-based kubeconfig auth (`aws eks get-token`).

In [`createHelmActionConfig`](https://github.com/meshery/meshkit/blob/master/utils/kubernetes/apply-helm-chart.go#L411), Meshkit sets [`kubeConfig.KubeConfig = /dev/null`](https://github.com/meshery/meshkit/blob/master/utils/kubernetes/apply-helm-chart.go#L429) and wires only `APIServer` + `BearerToken`. For exec-auth contexts, `rest.Config.ExecProvider` is populated but `rest.Config.BearerToken` is empty, so Helm is initialized without usable credentials and returns:

```sh
could not install Meshery Server: Kubernetes cluster unreachable: the server has asked for the client to provide credentials
```

This causes both [Meshery Server and Meshery Operator](https://github.com/meshery/meshery/blob/master/mesheryctl/internal/cli/root/system/start.go#L519-L550) installation to fail in the same run. The wrapped error can also misleadingly include “Chart could be invalid.”

### Expected Behavior
`ApplyHelmChart` should support exec-auth kubeconfigs (EKS/GKE/AKS exec plugin flows). **In this case EKS is only tested.**

When `ExecProvider` is present and `BearerToken` is empty, Helm auth setup should use kubeconfig/exec-based credential resolution (or equivalent dynamic token handling) instead of forcing `/dev/null` + static bearer token.

Result: chart installation should succeed with the same kubeconfig that already works with `helm`.

### Screenshots/Logs

`mesheryctl system start -p kubernetes` fails with the following error:

```sh
Error: failed to start Meshery
See https://docs.meshery.io/reference/mesheryctl/system for usage details
: could not install Meshery Server: Kubernetes cluster unreachable: the server has asked for the client to provide credentials | Short Description: Error applying helm chart | Probable Cause: Chart could be invalid | Suggested Remediation: Make sure to apply valid chart | Short Description: Error applying helm chart | Probable Cause: Chart could be invalid | Suggested Remediation: Make sure to apply valid chart

could not install meshery-operator: Kubernetes cluster unreachable: the server has asked for the client to provide credentials | Short Description: Error applying helm chart | Probable Cause: Chart could be invalid | Suggested Remediation: Make sure to apply valid chart | Short Description: Error applying helm chart | Probable Cause: Chart could be invalid | Suggested Remediation: Make sure to apply valid chart
```

### Environment

- **Host OS:** Mac
- **Platform:** Kubernetes (EKS)
- **Meshery Server Version:**  v1.0.11
- **Meshery Client Version:** unavailable

### To Reproduce
1. Updated kubeconfig for EKS:
`aws eks update-kubeconfig --region <region> --name <cluster-name>`

2. Ran Meshery install on Kubernetes platform:
`mesheryctl system start -p kubernetes --skip-browser`

### RCA: 

Detailed RCA can be found [here](https://docs.google.com/document/d/1xSacSEVmDa6YFRq4NaxXX__TnbXu_Zn9Hv2NW_suS74/edit?tab=t.0#heading=h.7lmje1zey82z)

---
### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)
- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/GvB8SudhEOoq3JOvoLaoMs)

reaper8055:
**Description**

This PR fixes #985

> [!NOTE]
>  There's a major shift in the original proposed implementation and the latest changes, please refer to this comment: https://github.com/meshery/meshkit/pull/986#issuecomment-5306466200

**Notes for Reviewers**

Added `useKubeconfigAuth := c.RestConfig.ExecProvider != nil && c.RestConfig.BearerToken == ""` to [`createHelmActionConfig`](https://github.com/meshery/meshkit/blob/master/utils/kubernetes/apply-helm-chart.go#L411)

1. `c.RestConfig.ExecProvider != nil` means auth came from an exec plugin (EKS style aws eks get-token, etc).
2. `c.RestConfig.BearerToken == ""` means there is no static token available in the rest.Config right now.

(1.) && (2.) means:

Cluster needs exec-plugin token retrieval and static bearer-token wiring is insufficient.
Therefore, use kubeconfig-based auth resolution path (so client-go can execute plugin), instead of forcing /dev/null and expecting bearer token to already exist.

## Summary by CodeRabbit

* **New Features**
  * Improved Kubernetes and Helm connectivity through reusable client configuration support.
  * Added support for preserving kubeconfig loaders and renewing exec-based credentials during requests.
  * Added fallback handling for clients configured directly with REST settings.

* **Bug Fixes**
  * Improved kubeconfig discovery across provided settings, in-cluster configuration, environment variables, and default paths.
  * Simplified Helm authentication setup while retaining secure REST client access.
  * Improved reliability when initializing Helm with SQL-based drivers.


leecalcote:
When `useKubeconfigAuth` is true, `kubeConfig.KubeConfig` is left unset. Client-go falls back to $KUBECONFIG on the user's system, which could point to a different cluster. Helm potentially silently installs Meshery into the wrong cluster. 

We'll probably want to serializing the in-memory kubeconfig (including the ExecProvider) to a temp file, appending it to tempFiles, and pointing kubeConfig.KubeConfig at that file so cluster/context stay bound to what the user provided.

As we work through this, let's be aware of the flows initiated via both Meshery CLI and Server.

leecalcote:
Test coverage 👀

leecalcote:
It'd be good to get our follow up issue for GKE, OKE, AKS, opened and referenced now.

reaper8055:
> When `useKubeconfigAuth` is true, `kubeConfig.KubeConfig` is left unset. Client-go falls back to $KUBECONFIG on the user's system, which could point to a different cluster. Helm potentially silently installs Meshery into the wrong cluster.

Yes this can happen if the user runs `mesheryctl system start -p kubernetes` without running `aws eks update-kubeconfig` first, but since the user would have run `aws eks update-kubeconfig --name [YOUR_CLUSTER_NAME] --region [YOUR_REGION]` as per this https://docs.meshery.io/installation/kubernetes/eks/#in-cluster-installation the current context for k8s cluster in the `$KUBECONFIG` would be pointing to the right cluster. 

Maybe, we can have a prompt saying:
```
meshery is going to install in <cluster-name> obtained from your current context in $KUBECONFIG, are you sure to proceed (y/n)? 
```

> We'll probably want to serializing the in-memory kubeconfig (including the ExecProvider) to a temp file, appending it to tempFiles, and pointing kubeConfig.KubeConfig at that file so cluster/context stay bound to what the user provided.

I'm assuming, we would want to somehow persist this file across start/stop commands? And since `$KUBECONFIG` is updated by `aws eks update-kubeconfig`, one way to do this is to read the current kubecontext and copy it to a temp file using:

```go
f, err := os.CreateTemp("/tmp", "meshery-eks-kubeconfig")
```

**OR**

Initiate the EKS flow using aws sdk which generates the in memory config and use client-go to write this config to a temp file instead of relying on user to run `aws update kube-config`. In this case, we will have to prompt the user to input the cluster-name and region for their EKS cluster similar to what `mesheryctl connection create` command does.

> As we work through this, let's be aware of the flows initiated via both Meshery CLI and Server.

ACK! 

The saving of current config depends on the remote provider, which happens in two paths:

* Kubeconfig -> k8s connection payload (+ credential secret)
  * addK8SConfig() calls provider.SaveK8sContext()
  * Remote path is RemoteProvider.SaveK8sContext()
  * It sends CredentialSecret: {"auth":..., "cluster":...} through SaveConnection()
  
* Post-login self-registration (second meshery connection)

reaper8055:
> Test coverage 👀

Hello @leecalcote do you want me to increase the test coverage for this package? The current test coverage looks low:

```sh
ok  github.com/meshery/meshkit/utils/kubernetes	1.188s          coverage: 7.9% of statements
    github.com/meshery/meshkit/utils/kubernetes/describe        coverage: 0.0% of statements
    github.com/meshery/meshkit/utils/kubernetes/expose	        coverage: 0.0% of statements
ok  github.com/meshery/meshkit/utils/kubernetes/kompose	1.564s	coverage: 75.4% of statements
```

This can be done but will require a fair bit of refactoring of logic especially to be able to mock behavior effectively.

lekaf974:
@reaper8055 just wanted to confirm this is ready for review ?

reaper8055:
> @reaper8055 just wanted to confirm this is ready for review ?

Yes @lekaf974 this is ready for review

utils/kubernetes/apply-helm-chart.go
```diff
@@ -54,11 +56,16 @@ const (
 	Latest = ">0.0.0-0"
 )
 
+// downloadLocaton is the location where downloaded helm charts
+// will be stored. os.TempDir will ensure that the path is cross
+// platform
+var downloadLocation = os.TempDir()
```
lekaf974: why moving this ?

utils/kubernetes/apply-helm-chart.go
```diff
@@ -421,66 +443,105 @@ func createHelmActionConfig(c *Client, cfg ApplyHelmChartConfig) (*action.Config
...
+		Cluster:  clusterName,
+		AuthInfo: authInfo,
+	}
+
+	// explicitly setting kube context may not be required
+	helmKubeConfig.CurrentContext = clusterName
+
+	configBytes, err := writeKubeConfig(*helmKubeConfig)
+	if err != nil {
+		return nil, cleanup, fmt.Errorf("failed to write kubeconfig %v", err)
```
lekaf974: please return a meshkit error

utils/kubernetes/apply-helm-chart.go
```diff
@@ -421,66 +443,105 @@ func createHelmActionConfig(c *Client, cfg ApplyHelmChartConfig) (*action.Config
...
 	}
 
-	return actionConfig, cleanup, nil
+	configFile, err := writeTempData(configBytes)
+	if err != nil {
+		return nil, cleanup, fmt.Errorf("failed to get kubeconfig file %v", err)
```
lekaf974: same meshkit error needed

lekaf974 (changes requested):
small comments LGTM otherwise

utils/kubernetes/apply-helm-chart.go
reaper8055: Ohh yes, this is not needed to be separate, I added additional global variables for test below and forgot to move this.

utils/kubernetes/apply-helm-chart.go
reaper8055: Just one small clarification, in the `func (c *Client) ApplyHelmChart(cfg ApplyHelmChartConfig) error` function, the errors returned from `setupKubeConfig` call
already become a MeshKit error through [`ErrApplyHelmChart(err)`](https://github.com/meshery/meshkit/blob/master/utils/kubernetes/error.go#L83). 

Do we still need to make the errors in `setupKubeConfig` function meshkit errors? (unexported functions?). Imo if I were to return meshkit error from the `setupKubeConfig` function then we get a MeshKit error wrapped into another MeshKit error which makes the final error noisy and can hide the more precise code, no?

utils/kubernetes/apply-helm-chart.go
reaper8055: Just one small clarification, in the `func (c *Client) ApplyHelmChart(cfg ApplyHelmChartConfig) error` function, the errors returned from `setupKubeConfig` call
already become a MeshKit error through [`ErrApplyHelmChart(err)`](https://github.com/meshery/meshkit/blob/master/utils/kubernetes/error.go#L83). 

Do we still need to make the errors in `setupKubeConfig` function meshkit errors? (unexported functions?). Imo if I were to return meshkit error from the `setupKubeConfig` function then we get a MeshKit error wrapped into another MeshKit error which makes the final error noisy and can hide the more precise code, no?

utils/kubernetes/apply-helm-chart.go
reaper8055: fixed in latest commit.

utils/kubernetes/apply-helm-chart.go
lekaf974: You are right we can skip for this time but I think the best would have been have a meshkit error (eg. ErrSetupKubeConfig) within setupKubeConfig and return it as is in ApplyHelmChart this way it is easier to identify from where the error is trigger and why, that say it can be done in another step

utils/kubernetes/apply-helm-chart.go
lekaf974: same

lekaf974 (approved):

PragalvaXFREZ (approved):
LGTM

utils/kubernetes/apply-helm-chart.go
```diff
@@ -421,66 +444,105 @@ func createHelmActionConfig(c *Client, cfg ApplyHelmChartConfig) (*action.Config
...
+		AuthProvider: c.RestConfig.AuthProvider,
+	}
+
+	helmKubeConfig.Contexts[clusterName] = &clientcmdapi.Context{
+		Cluster:  clusterName,
+		AuthInfo: authInfo,
 	}
 
-	return actionConfig, cleanup, nil
+	// explicitly setting kube context may not be required
```
PragalvaXFREZ: i find this comment a bit conflicting..

PragalvaXFREZ: Rationale: I think client-go falls back to current-context and won't auto-select even with one context, so dropping `helmKubeConfig.CurrentContext = clusterName` resolves to an empty server. Therefore I think it's needed.

leecalcote:
We're a couple of months into this one. I have my 🤞 for it.

utils/kubernetes/apply-helm-chart.go
reaper8055: Yeah, that's my bad. It was just for me to review if this was required or not and I forgot to remove the comment. Thanks for catching this else this would have been confusing!

reaper8055:
> We're a couple of months into this one. I have my 🤞 for it.

Resoling the comments and updating the PR as I write this. Will get this done!

utils/kubernetes/apply-helm-chart.go
reaper8055: Resolved in latest commit.

aabidsofi19:
@reaper8055  regarding error handling are we bubbling up errors like missing the exec binary , failure to exec etc to users

reaper8055:
> @reaper8055 regarding error handling are we bubbling up errors like missing the exec binary , failure to exec etc to users

the kubeConfig ~api~ backend that does invoke the exec call does that implicitly so we don't have to do it.

utils/kubernetes/apply-helm-chart.go
```diff
@@ -421,57 +419,64 @@ func createHelmActionConfig(c *Client, cfg ApplyHelmChartConfig) (*action.Config
...
+		kubeConfig.Insecure = &c.RestConfig.Insecure
+
+		// Set username and password for basic auth if available
+		if c.RestConfig.Username != "" {
+			kubeConfig.Username = &c.RestConfig.Username
+		}
+		if c.RestConfig.Password != "" {
+			kubeConfig.Password = &c.RestConfig.Password
+		}
+
```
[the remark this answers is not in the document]
ritzorama: Please reply to each review comment. Either show how you incorporated the feedback or explain why you're rejecting it.

utils/kubernetes/apply-helm-chart.go
```diff
@@ -408,96 +407,16 @@ func checkIfInstallable(ch *chart.Chart) error {
 }
 
 // createHelmActionConfig generates the actionConfig with the appropriate defaults
-func createHelmActionConfig(c *Client, cfg ApplyHelmChartConfig) (*action.Configuration, func(), error) {
+func (c *Client) createHelmActionConfig(cfg ApplyHelmChartConfig, restClientGetter genericclioptions.RESTClientGetter) (*action.Configuration, error) {
 	// Set the environment variable needed by the Init methods
 	_ = os.Setenv("HELM_DRIVER_SQL_CONNECTION_STRING", cfg.SQLConnectionString)
```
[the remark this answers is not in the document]
reaper8055: Done. 

The SQL connection string is now set only when the SQL Helm driver is selected and a non-empty connection string is provided.
Since Helm only accepts this value through `HELM_DRIVER_SQL_CONNECTION_STRING`, the set/init/restore sequence is protected by a mutex. The previous environment value is restored or the variable is removed immediately after Helm initializes its storage driver. `os.Setenv` errors are also propagated.

This prevents concurrent Helm initializations from reading each other’s connection strings and avoids leaving the credential available to subsequently executed credential plugins.

utils/kubernetes/apply-helm-chart.go
```diff
@@ -643,7 +563,7 @@ func (helmEntries HelmEntries) GetEntryWithAppVersion(entry, appVersion string)
 	return HelmEntryMetadata{}, false
 }
 
-// GetEntryWithAppVersion takes in the entry name and the appversion and returns the corresponding
+// GetEntryWithChartVersion takes in the entry name and the appversion and returns the corresponding
```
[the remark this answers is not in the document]
reaper8055: Fixed. The GetEntryWithChartVersion comment now describes the parameter as the chart version rather than the app version

utils/kubernetes/client-config-getter.go
```diff
@@ -0,0 +1,101 @@
...
+		CertificateAuthority:     c.restConfig.CAFile,
+		CertificateAuthorityData: c.restConfig.CAData,
+		DisableCompression:       c.restConfig.DisableCompression,
+	}
+	config.Contexts[connectionName] = &clientcmdapi.Context{
+		Cluster: connectionName,
+	}
+	config.CurrentContext = connectionName
+	return *config, nil
+}
```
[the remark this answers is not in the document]
reaper8055: Fixed. The synthetic configuration now includes an AuthInfo entry and links the generated context to it.

It preserves the serializable authentication and connection fields from rest.Config, including bearer token/file, client certificate/key, username/password, auth provider, exec provider, impersonation configuration, and resolved proxy URL.
The generated cluster and authentication entries are deep-copied so consumers cannot mutate the authoritative REST configuration through the returned raw configuration.

utils/kubernetes/client.go
```diff
@@ -12,47 +12,66 @@ import (
...
 		if err != nil {
-			return nil, err
+			return nil, nil, err
 		}
-		if config, err = clientcmd.RESTConfigFromKubeConfig(cfgFile); err == nil {
-			return config, nil
+		if config, loader, err = loadClientConfigFromKubeconfig(cfgFile); err == nil {
+			return config, loader, nil
 		}
 	}
```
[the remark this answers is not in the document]
reaper8055: The error shadowing has been fixed by assigning to the existing named error variable, so failures from the explicitly selected kubeconfig are now returned correctly.

I intentionally did not continue to ~/.kube/config when $KUBECONFIG is explicitly set but invalid. 

$KUBECONFIG expresses the user’s intended cluster, and silently falling back could run operations against a different cluster. This also matches client-go’s default loading precedence, which uses $KUBECONFIG when present and only uses the default home file otherwise.

The branch therefore now fails closed while preserving the actual processing or loading error.

reaper8055:
Additional notes/changes that code-rabbit did not catch in latest commit:

1. Cache the fallback REST client getter

clientConfigRESTClientGetter does not own or retain the discovery cache; ToDiscoveryClient constructs a new memory-backed discovery client on every invocation. Caching only the getter would therefore not share discovery results.
It would also freeze the first snapshot of a directly constructed client’s RestConfig and require synchronization around newly mutable client state. Given the negligible allocation and absence of the stated cache benefit, retaining the current behavior is safer.

2. Preserve caller-configured QPS and burst

configureRESTConfig now applies MeshKit’s 50 QPS and 100 burst defaults only when the corresponding fields are zero. Explicit caller-provided values are preserved.
Tests now cover both default application and preservation of direct-client rate limits.

3. Anchor the helper-process test pattern

Fixed. The helper subprocess now uses -test.run=^TestExecCredentialHelperProcess$, ensuring that only the intended helper test runs and preventing unrelated test output from corrupting the exec credential JSON.

reaper8055:
### Update on the implemented approach

The original fix correctly identified why EKS installation was failing: Helm was receiving a reconstructed Kubernetes configuration that did not preserve exec-based authentication such as `aws eks get-token`.

The first implementation addressed this inside `ApplyHelmChart` by:

- Detecting exec authentication through `ExecProvider`.
- Maintaining separate paths for exec and static credentials.
- Reconstructing a synthetic kubeconfig.
- Writing kubeconfig, CA, certificate, and key data to temporary files.
- Passing the resulting configuration to Helm and cleaning up afterward.
- Later preferring the original kubeconfig loader when one was available.

This worked, but it placed authentication responsibility in the wrong layer. Helm should consume an established Kubernetes connection; it should not classify authentication mechanisms or reconstruct kubeconfigs.

### What changed

Kubernetes configuration is now established during `kubernetes.Client` initialization:

1. For kubeconfig-backed connections, MeshKit retains the original `clientcmd.ClientConfig`.
2. For in-cluster or directly constructed clients, MeshKit provides a `rest.Config`-backed adapter.
3. A shared `RESTClientGetter` is initialized from that source.
4. The typed, dynamic, discovery, and Helm clients all derive their configuration from the same connection source.
5. Helm now receives this getter directly and contains no authentication-specific branching.

As a result, client-go remains responsible for executing and renewing credentials. MeshKit no longer needs to explicitly check whether authentication uses exec, bearer tokens, certificates, or another supported mechanism.

The following Helm-specific workaround code has been removed:

- `setupKubeConfig`
- Exec-versus-static authentication classification
- Synthetic kubeconfig serialization
- Temporary kubeconfig, CA, certificate, and key files
- Associated cleanup callbacks

### Additional hardening

The latest changes also:

- Preserve serializable authentication, impersonation, TLS, and proxy information in `RawConfig`.
- Return deep copies so consumers cannot mutate the authoritative `rest.Config`.
- Fail closed when an explicitly configured `$KUBECONFIG` is invalid instead of potentially targeting the default cluster.
- Scope and restore Helm’s SQL connection-string environment variable.
- Prevent duplicate `ErrApplyHelmChart` wrapping.
- Preserve caller-configured QPS and burst limits.
- Add focused regression coverage for exec credential renewal and the new configuration adapters.

The updated implementation has been tested successfully using:

```shell
mesheryctl system start -p kubernetes
```

The main architectural change is that authentication is now initialized once at the Kubernetes connection boundary and reused by all consumers, instead of being reconstructed specifically for Helm.

utils/kubernetes/apply-helm-chart.go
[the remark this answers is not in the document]
reaper8055: That implementation has since been replaced rather than patched. Helm no longer builds ConfigFlags, conditionally leaves KubeConfig unset, or separately applies the API server and TLS fields.

Kubernetes configuration is now resolved during kubernetes.Client initialization:

- For kubeconfig-backed clients, MeshKit retains the processed clientcmd.ClientConfig and creates the shared RESTClientGetter from it.
- The same configuration source supplies the API server, TLS settings, selected context, and authentication—including exec-based authentication.
- ApplyHelmChart passes that initialized getter directly to Helm’s action.Configuration.Init.
- Direct and in-cluster rest.Config clients use an adapter that preserves the cluster and authentication information.

Therefore, Helm no longer falls back through an unset ConfigFlags.KubeConfig, and it cannot independently select a different local cluster through this path.

lekaf974:
@reaper8055 looks good to merge anything else on your side to do ?

reaper8055:
> @reaper8055 looks good to merge anything else on your side to do ?

No, nothing else. I have resolved and addressed all comments. And the implementation is also tested against a aws cluster in my personal account.

PR #1007, opened 2026-05-13
[retry] Add centralized retry/backoff abstraction

### LINKED ISSUE #1008: Feature Request: Centralized Retry/Backoff Helper in MeshKit
Current Situation
Retry and backoff logic across Meshery components is currently implemented in an ad hoc, per-call-site manner. This can lead to:

duplicated retry implementations
inconsistent retry policies and backoff strategies
fragmented observability/logging behavior
harder long-term maintenance
During discussion in a related PR, it was noted that Meshery does not yet have a standardized retry/backoff abstraction.

Existing Precedent
A reusable retry pattern already exists in parts of the ecosystem, for example in meshery-adapter-library, which uses cenkalti/backoff/v4 in adapter/meshmodel.go.

This suggests an opportunity to move toward a shared abstraction in MeshKit, which is already a common dependency across Meshery components.

Proposal
Introduce a centralized retry/backoff helper in MeshKit that can be reused across Meshery projects.

Potential goals:

standardized retry behavior
configurable backoff strategies
reusable helper APIs instead of hand-rolled retry loops
improved consistency across components
centralized observability/logging hooks
easier future maintenance and tuning
Possible Scope
Initial implementation could focus on:

exponential backoff support
configurable retry counts/durations
retry conditions based on error types
context-aware cancellation support
lightweight and composable API surface
Non-Goals (Initial Version)
workflow orchestration
distributed retry coordination
persistence/stateful retry queues
Related Discussion
This issue originates from discussion around introducing retry logic in a component-specific PR https://github.com/meshery/meshery/pull/19254 , https://github.com/meshery/meshery/pull/19227 , where it was suggested that the functionality should instead be abstracted into MeshKit for reuse across the ecosystem.

Pranav-d33:
cc @leecalcote   Introduces retry/ — a context-aware, exponential-backoff retry package wrapping cenkalti/backoff/v4. Provides a reusable abstraction to replace ad-hoc retry loops across Meshery components.

Changes:
- Add retry/{retry,options,errors,retry_test}.go
  - Do(ctx, op, ...opts) with production-safe defaults (500 ms initial, x1.5 growth, +/-30% jitter, 2 min cap)
  - Permanent(err) escape hatch for non-retryable errors
  - IsPermanent(err) using errors.As for proper chain unwrapping
  - WithLogNotifier bridges retries to MeshKit logger.Handler
  - 15 table-driven tests, race-detector clean
- Promote cenkalti/backoff/v4 v4.3.0 from indirect to direct in go.mod
fixes #1008


Pranav-d33:
cc @leecalcote  Wanted to flag this is ready for review when you get a chance — happy to walk through the approach or adjust based on feedback.

retry/error.go
```diff
@@ -0,0 +1,34 @@
+package retry
+
+import (
+	"errors"
+
+	"github.com/cenkalti/backoff/v4"
```
icoderarely: use latest version, refer the repo: https://github.com/cenkalti/backoff

retry/retry.go
```diff
@@ -0,0 +1,97 @@
...
+//	    if resp.StatusCode == http.StatusBadRequest {
+//	        return retry.Permanent(fmt.Errorf("bad request: %d", resp.StatusCode))
+//	    }
+//	    return nil
+//	}, retry.WithMaxAttempts(5), retry.WithLogNotifier(log))
+//
+// # Error classification
+//
+//   - Network errors, HTTP 5xx, HTTP 429  → return err (transient)
+//   - HTTP 4xx (except 429), auth, decode → return retry.Permanent(err)
```
icoderarely: lots of comments, if the codebase has such comments everywhere then feel free to use else try to keep it clean + simple to read but modular

retry/retry_test.go
```diff
@@ -0,0 +1,360 @@
...
+func countingOp(count *atomic.Int64, err error) retry.Operation {
+	return func(ctx context.Context) error {
+		count.Add(1)
+		return err
+	}
+}
+
+// ─── tests ──────────────────────────────────────────────────────────────────
+
+func TestDo_SucceedsFirstAttempt(t *testing.T) {
```
icoderarely: check the naming convention of tests in the repo and try to match those

icoderarely (changes requested):

icoderarely (commented):
LGTM, might need a maintainers eye on it now for this to be merged if this is a fix they want. Good work @Pranav-d33

Pranav-d33:
@icoderarely Can you point me to the right person to ping ???

icoderarely:
**hey, did you not create an issue or forgot to link this pr to an issue you are working on?** 

and the _one who assigned you_ might be the first person you could tag. 

Also have some patience, it's a big project, alot of work is happening in parallel and maintainers/members need time too to get back to each. So best bet would be to tag, wait for a day, if there is no response msg this on slack community that you need this pr to be reviewed/merged.

Pranav-d33:
The issue is in Meshery/meshery and lee was the one who assigned it to me, thats why i pinged him in slack... 
https://github.com/meshery/meshery/issues/19275

icoderarely:
oh we gotta tag someone else then (some exceptions mate)

icoderarely:
@Omkar-Ugal who could be the right person to tag?

Omkar-Ugal:
please link the pr with the issue

Pranav-d33:
> please link the pr with the issue

it is linked, https://github.com/meshery/meshery/issues/19275 ,  the issue is in meshery actually, do i create a new issue in meshkit as well?

go.mod
```diff
@@ -100,7 +101,6 @@ require (
 	github.com/aws/smithy-go v1.24.0 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/blang/semver/v4 v4.0.0 // indirect
-	github.com/cenkalti/backoff/v5 v5.0.3 // indirect
```
matrixkavi, line 103: Do we need this change?

matrixkavi (changes requested):
The change for go.mod is not required.

 While majority of the code looks ok to me, I was wondering what would happen if a caller called` WithMultiplier` or` WithJitter` with negative values for instance. Isn't it good to validate and set the values to default rather than blindly setting them in options.go?

Pranav-d33:
Added validation guards:

* `WithMultiplier`: values `<= 0` now fall back to `DefaultMultiplier (1.5)`
* `WithJitter`: values `< 0` or `> 1` now fall back to `DefaultRandomizationFactor (0.3)`

Also reverted the unnecessary `go.mod` change.

retry/retry.go
```diff
@@ -0,0 +1,47 @@
...
+		o(&cfg)
+	}
+
+	retryOpts := []backoff.RetryOption{
+		backoff.WithBackOff(buildBackOff(cfg)),
+		backoff.WithMaxElapsedTime(cfg.MaxElapsedTime),
+		backoff.WithNotify(cfg.Notifier),
+	}
+	if cfg.MaxAttempts > 0 {
+		retryOpts = append(retryOpts, backoff.WithMaxTries(uint(cfg.MaxAttempts)))
```
lekaf974: Why cast uint is necessary ?

lekaf974 (commented):
Just a quick commen, LGTM otherwise

retry/retry.go
Pranav-d33: cfg.MaxAttempts is uint64 and backoff.WithMaxTries() takes uint. 
Could be avoided by changing MaxAttempts to uint in the config, but that couples the config type to the backoff library's signature — not worth it for eliminating one cast.

Pranav-d33:
@matrixkavi , Let me know if any changes are needed before merging.

retry/retry.go
lekaf974: IMO we are already coupled with backoff library, do we really need unit64 in our implementation ? I mean what is the advantage of one over the other?

Pranav-d33: Good point, that makes sense. Since we’re already directly coupled to the backoff library APIs and semantics, using `uint64` here doesn’t provide any real abstraction benefit over `uint`.

I’ve updated `MaxAttempts` and `WithMaxAttempts()` to use `uint` instead, which removes the unnecessary cast at the call site and keeps the implementation cleaner and more consistent with `backoff.WithMaxTries()`.

icoderarely (approved):
LGTM for 
> retry: drop unnecessary uint64->uint cast by aligning Config.MaxAttempts type with backoff library

lekaf974 (dismissed):
LGTM

leecalcote:
There is the smallest amount of discourse on this topic in the community Slack - https://mesheryio.slack.com/archives/CFGG6U10E/p1779755449286469?thread_ts=1776923231.353449&cid=CFGG6U10E

leecalcote:
@Pranav-d33 this _might_ be something that we can wire into the remote provider availability tests that occur concurrently onload of the Provider UI.

Pranav-d33:
## MeshKit Retry Abstraction Test Matrix

### Category 1: Basic Success Paths

| Scenario | What it tests | Result |
|---|---|---|
| `TestRetrySucceedsFirstAttempt` | Operation succeeds on the first call with no retries. | PASS (0.00s) — operation called exactly once |
| `TestRetrySucceedsAfterTransientErrors` | Operation fails 3 times and succeeds on the 4th attempt. | PASS (0.01s) — operation called exactly 4 times |
| `TestRetryDefaultsAreApplied` | Verifies default retry settings are applied when no options are provided. | PASS (0.36s) — at least 2 calls made |

### Category 2: Permanent Error Handling

| Scenario | What it tests | Result |
|---|---|---|
| `TestRetryPermanentErrorStopsImmediately` | Permanent errors stop retries immediately. | PASS (0.00s) — exactly 1 call |
| `TestIsPermanentReturnsFalseForTransient` | Transient errors are not classified as permanent. | PASS (0.00s) |
| `TestIsPermanentReturnsTrueForPermanentWrapped` | Wrapped permanent errors are correctly detected. | PASS (0.00s) |
| `TestIsPermanentHandlesDoublyWrappedErrors` | Nested wrapped permanent errors still unwrap correctly. | PASS (0.00s) |

### Category 3: Context Cancellation

| Scenario | What it tests | Result |
|---|---|---|
| `TestRetryContextCancellationStopsLoop` | Retry loop stops when context is cancelled mid-backoff. | PASS (0.01s) |
| `TestRetryContextAlreadyCancelledBeforeFirstAttempt` | Retry exits immediately if context is already cancelled. | PASS (0.00s) |

### Category 4: Budget Enforcement

| Scenario | What it tests | Result |
|---|---|---|
| `TestRetryMaxAttemptsEnforced` | Retry count respects `WithMaxAttempts(4)`. | PASS (0.01s) — exactly 4 calls |
| `TestRetryMaxElapsedTimeEnforced` | Retry loop respects wall-clock elapsed time budget. | PASS (0.08s) |
| `TestRetryZeroMaxAttemptsMeansUnlimited` | Unlimited retries stop only via elapsed time budget. | PASS (0.04s) |
| `TestRetryWithMaxAttemptsOneNoRetry` | `WithMaxAttempts(1)` performs exactly one attempt. | PASS (0.00s) |

### Category 5: Notifier Callback Behavior

| Scenario | What it tests | Result |
|---|---|---|
| `TestRetryNotifierCalledOnEachRetry` | Notifier invoked once per transient retry failure. | PASS (0.01s) |
| `TestRetryNotifierNotCalledOnImmediateSuccess` | Notifier is not called on immediate success. | PASS (0.00s) |
| `TestRetryNotifierNotCalledOnPermanentError` | Notifier is not called for permanent errors. | PASS (0.00s) |

### Summary

- Total scenarios executed: **16**
- Passed: **16/16**
- Total runtime: **0.36s**
- Coverage includes:
  - transient retry behavior
  - exponential backoff defaults
  - permanent error semantics
  - context cancellation
  - retry budget enforcement
  - notifier callback behavior
  - wrapped error handling

retry/retry.go
```diff
@@ -0,0 +1,96 @@
...
+//   - the configured MaxAttempts or MaxElapsedTime budget is exhausted
+//
+// All retry configuration is supplied via the variadic opts. When no opts are
+// given, [defaultConfig] values are used (500 ms initial, ×1.5 growth, 30 %
+// jitter, 2-min elapsed cap). Individual knobs are overridden with the With*
+// helpers defined in options.go.
+//
+// Do is safe for concurrent use; each call creates its own backoff state.
+func Do(ctx context.Context, op Operation, opts ...Option) error {
+	cfg := defaultConfig()
```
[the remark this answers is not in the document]
Pranav-d33: resolved

retry/options.go
```diff
@@ -0,0 +1,88 @@
+package retry
```
reaper8055, line 1: The retry package needs an explicit error-classification hook.

Right now `retry.Do` retries every non-nil error unless the caller manually wraps it with `retry.Permanent`. That behavior is easy to miss, and it puts too much semantic responsibility on every call site. In practice, callers have to remember to classify errors correctly every time they use this helper, otherwise permanent failures get treated as transient ones.

For example, a `404 Not Found` from an HTTP endpoint is usually not transient. Retrying it several times just adds latency and load; it will likely keep returning `404`.

A better API would let the caller provide a classifier:

```go
retry.Do(ctx, op,
    retry.WithErrorClassifier(func(err error) retry.ErrorDecision {
        var status statusError
        if errors.As(err, &status) {
            switch {
            case status.Code == http.StatusTooManyRequests:
                return retry.DecisionRetry
            case status.Code >= 500:
                return retry.DecisionRetry
            default:
                return retry.DecisionStop
            }
        }

        return retry.DecisionRetry
    }),
)
```

**OR**

```go
retry.Do(ctx, op,
    retry.WithErrorClassifier(func(err error) retry.ErrorDecision {
        if isHTTPStatus(err, http.StatusNotFound) {
            return retry.DecisionStop
        }
        return retry.DecisionRetry
    }),
)
```


This makes retry policy explicit at the bounded context where the error semantics are known. The retry package should own the retry loop mechanics, but the caller should own the domain-specific decision of whether an error is transient or terminal.

Without this, the abstraction is technically usable but too footgun-prone: a call site can accidentally retry auth failures, validation errors, malformed requests, or unsupported endpoint responses just because they were returned as ordinary error values.

retry/options.go
```diff
@@ -0,0 +1,88 @@
...
+// WithJitter overrides randomization factor (range: 0.0-1.0). Do not set to 0.0 in production.
+func WithJitter(f float64) Option {
+	return func(c *Config) {
+		if f < 0 || f > 1 {
+			c.RandomizationFactor = DefaultRandomizationFactor
+			return
+		}
+		c.RandomizationFactor = f
+	}
+}
```
reaper8055, line 89: The retry configuration should be validated before starting the retry loop.

Right now options like `WithInitialInterval` and `WithMaxInterval` accept zero or negative durations, and `WithMultiplier` / `WithJitter` do not reject invalid floating-point values like `NaN` or `Inf`.

That makes it possible to accidentally create a pathological retry policy. For example:

```go
retry.Do(ctx, op,
    retry.WithInitialInterval(0),
    retry.WithMaxAttempts(0),
    retry.WithMaxElapsedTime(0),
)
```

This can become an unbounded tight retry loop: no attempt cap, no elapsed-time cap, and no meaningful delay between attempts. During an outage, that turns a dependency failure into local CPU pressure and unnecessary downstream load.

The retry package should fail fast on invalid config before calling into the backoff library. At minimum, I’d expect validation for:

- `InitialInterval > 0`
- `MaxInterval > 0`
- `MaxInterval >= InitialInterval`
- `MaxElapsedTime >= 0`
- `Multiplier` is finite and probably `>= 1`
- `Jitter` is finite and within `[0, 1]`

Also, silently replacing invalid multiplier/jitter values with defaults hides caller bugs. Returning a config error is safer and easier to debug.

retry/error.go
```diff
@@ -0,0 +1,22 @@
+package retry
+
+import (
+	"errors"
+
+	"github.com/cenkalti/backoff/v5"
+)
+
+type PermanentError = backoff.PermanentError
```
reaper8055: The retry package should avoid exposing the backing library as part of MeshKit’s public API.

Right now this line leaks `cenkalti/backoff` directly:

```go
type PermanentError = backoff.PermanentError
```

That means downstream consumers can start depending on `*retry.PermanentError`, which is actually `*backoff.PermanentError` (line:20 in same file) under the alias. Once that happens, MeshKit can no longer replace or wrap the backoff implementation without creating a breaking API change.

The package already exposes the right behavioral surface:

```go
retry.Permanent(err)
retry.IsPermanent(err)
```

Those are enough for callers to mark and inspect terminal errors without coupling themselves to the third-party implementation.

I’d remove the exported type alias and keep the dependency behind MeshKit’s own API boundary. The retry package should act as an anti-corruption layer around `cenkalti/backoff`, not re-export its internals.

retry/retry.go
```diff
@@ -0,0 +1,47 @@
+package retry
```
reaper8055, line 1: The retry package should include a clear HTTP usage example.

This abstraction is likely to be used around outbound HTTP calls, so the docs should show the important distinction between retry budget and per-attempt timeout.

`MaxElapsedTime` only limits the retry loop after an operation returns. It does not magically interrupt an operation that is stuck inside a network call. For HTTP, each attempt still needs to be bound with `http.Client.Timeout` and/or a request created with `http.NewRequestWithContext`.

A good example would look roughly like this:

```go
client := &http.Client{Timeout: 3 * time.Second}

err := retry.Do(ctx, func(ctx context.Context) error {
    req, err := http.NewRequestWithContext(ctx, http.MethodGet, endpoint, nil)
    if err != nil {
        return retry.Permanent(err)
    }

    resp, err := client.Do(req)
    if err != nil {
        return err
    }
    defer resp.Body.Close()

    switch {
    case resp.StatusCode == http.StatusOK:
        return nil
    case resp.StatusCode == http.StatusTooManyRequests || resp.StatusCode >= 500:
        return fmt.Errorf("transient response: %s", resp.Status)
    default:
        return retry.Permanent(fmt.Errorf("non-retryable response: %s", resp.Status))
    }
},
    retry.WithMaxAttempts(3),
    retry.WithInitialInterval(time.Second),
    retry.WithMaxElapsedTime(10*time.Second),
)
```

This matters because otherwise callers may assume `WithMaxElapsedTime` protects them from a hung HTTP request. It does not. The retry package should document the correct pattern so users apply both boundaries: one for each attempt, and one for the overall retry loop.

reaper8055 (commented):
I have reviewed the changes, please check the comments.

retry/error.go
Pranav-d33: Removed PermanentError = backoff.PermanentError type alias to keep cenkalti/backoff behind MeshKit's API boundary. Added ErrorDecision (DecisionRetry/DecisionStop) and ErrorClassifier types.

retry/options.go
Pranav-d33: Added ErrorClassifier field to Config and WithErrorClassifier() option. Removed silent-default fallback from WithMultiplier/WithJitter (validation is now centralized).

retry/retry.go
Pranav-d33: Added validateConfig() that fails fast on invalid config (InitialInterval ≤ 0, MaxInterval ≤ 0, MaxInterval < InitialInterval, MaxElapsedTime < 0, Multiplier NaN/Inf/<1, Jitter NaN/out-of-range). Integrated classifier into Do() — runs on every non-nil error after op, converts DecisionStop to PermanentError; skips the classifier for explicitly wrapped Permanent() errors.

Pranav-d33:
@reaper8055 ready for re-review.

retry/retry.go
```diff
@@ -0,0 +1,98 @@
...
+		retryOpts = append(retryOpts, backoff.WithMaxTries(cfg.MaxAttempts))
+	}
+
+	_, err := backoff.Retry(ctx, func() (struct{}, error) {
+		return struct{}{}, apply(ctx)
+	}, retryOpts...)
+	return err
+}
+
+func validateConfig(cfg Config) error {
```
reaper8055, line 70: The config validation is a good addition, but the returned errors are currently string-only.

Right now callers can only distinguish invalid retry configuration from an operation failure by inspecting the error message. That is brittle. Since `retry.Do` can now fail before the operation even runs, callers should have a reliable way to detect that class of failure.

I’d suggest exposing a sentinel error, for example:

```go
var ErrInvalidConfig = errors.New("retry: invalid config")
```

Then wrap it from validateConfig:
```go
return fmt.Errorf("%w: InitialInterval must be > 0, got %v", ErrInvalidConfig, cfg.InitialInterval)
```

This lets consumers write:
```go
if errors.Is(err, retry.ErrInvalidConfig) {
    // fail startup / reject bad config
}
```

retry/retry_test.go
```diff
@@ -0,0 +1,686 @@
...
+		t.Fatalf("expected MaxElapsedTime validation error, got %v", err)
+	}
+}
+
+// ExampleDo demonstrates idiomatic HTTP usage with retry budget and per-attempt timeout.
+//
+// MaxElapsedTime limits the retry loop but does NOT interrupt an in-flight HTTP
+// request. Always pair it with http.Client.Timeout (or NewRequestWithContext) so
+// each attempt has its own deadline.
+func ExampleDo() {
```
reaper8055, line 662: The classifier API has good tests, but it also needs a compiled documentation example.

`WithErrorClassifier` is now an important part of the public contract because it lets callers separate retry-loop mechanics from domain-specific error classification. However, the current `ExampleDo` demonstrates `retry.Permanent` inside the operation, not the classifier path.

I’d add an example like `ExampleWithErrorClassifier` that shows the intended usage:

```go
err := retry.Do(ctx, op,
    retry.WithErrorClassifier(func(err error) retry.ErrorDecision {
        var status statusError
        if errors.As(err, &status) {
            if status.Code >= 500 || status.Code == http.StatusTooManyRequests {
                return retry.DecisionRetry
            }
            return retry.DecisionStop
        }
        return retry.DecisionRetry
    }),
)
```

This matters because otherwise users may keep embedding permanent/transient decisions inside every operation body. The cleaner contract is: the operation returns the domain error, and the classifier decides whether that error is retryable.

A compiled example also protects the public API from drift because go test will fail if the documented usage stops compiling.

**IMPORTANT**: check with @leecalcote / @lekaf974 on repo policy to where to include code examples for such use case!

retry/_demo/main.go
```diff
@@ -0,0 +1,175 @@
+package main
```
reaper8055: IMPORTANT: check with @leecalcote / @lekaf974 on repo policy to where to include code examples / demos for such use case!

reaper8055 (commented):
Everything looks good here apart from small things, check the comments.

retry/_demo/main.go
lekaf974, line 1: this was not part of my previous approval, I don't see any value to add this in meshkit

lekaf974 (commented):
Not sure that demo is needed

retry/retry.go
```diff
@@ -0,0 +1,98 @@
...
+type Operation func(ctx context.Context) error
+
+// Do executes op with exponential backoff until success, permanent error,
+// context cancellation, or budget exhaustion. Config via opts (default:
+// 500ms initial, 1.5x growth, 30% jitter, 2min cap).
+//
+// When a ErrorClassifier is configured via WithErrorClassifier, every non-nil
+// error from op (except those explicitly wrapped with Permanent) is passed to
+// the classifier before the retry decision is made.
+func Do(ctx context.Context, op Operation, opts ...Option) error {
```
lekaf974, line 21: All returned errors must be meshkit errors

lekaf974 (changes requested):
Please return meshkit errors from retry package

retry/_demo/main.go
Pranav-d33: Thanks for the feedback. I've removed the standalone _demo approach and added a compiled ExampleWithErrorClassifier in retry/retry_test.go instead.

retry/retry.go
Pranav-d33: Thanks for the feedback. I'll update the retry package so that all returned errors are MeshKit errors and adjust the validation/error paths accordingly.

banana-three-join (commented):
Nothing major really sticks out to me. I think it looks good!

PR #1066, opened 2026-07-19
refactor: rename models/meshmodel to models/registry, add compat shim…

Ayush-kr-giga:
## Description

Renames the `models/meshmodel` package to `models/registry` in meshkit, mirroring the `/api/meshmodel(s)` → `/api/registry` rename already underway in `meshery` meshery/meshery#20632 . This is the meshkit portion of the broader multi-repo capabilities registry rename effort.

## What changed

- `models/meshmodel/core` → `models/registry/core`
- `models/meshmodel/entity` → `models/registry/entity`
- `models/meshmodel/registry` → `models/registry/manager` (renamed to avoid a `registry/registry` stutter; matches the existing `RegistryManager` type)
- Updated all internal import paths across meshkit referencing the old `models/meshmodel/...` paths (`utils/utils.go`, `models/registration/*.go`, `models/patterns/pattern.go`, `registry/model.go`, `registry/component.go`, and files within the moved packages themselves)
- Added deprecated type aliases for two exported types that may be referenced by downstream consumers:
  - `MeshModelHostsWithEntitySummary` → alias for `RegistryHostsWithEntitySummary`
  - `MeshModelRegistrantData` → alias for `RegistryRegistrantData`
- Added a compatibility shim package at the old `models/meshmodel/entity` path (see Notes for Reviewers)

## Notes for Reviewers

[`meshery/schemas@v1.3.26(https://github.com/meshery/schemas/releases)`](https://github.com/meshery/schemas/releases) (meshkit's currently pinned dependency) has several helper files (`component_helper.go`, `model_helper.go`, `category_helper.go`, etc.) that import `meshkit/models/meshmodel/entity` directly, by convention, since those types implement meshkit's `entity.Entity` interface. Since meshkit is the main module in this build, Go resolves that import against meshkit's own local source tree — so removing the old path outright breaks the build for any consumer on this schemas version.

To keep things compiling without waiting on a `schemas` update, this PR adds a small compatibility shim at `models/meshmodel/entity` containing pure type aliases (`type X = Y`) to the real types now in `models/registry/entity`. Since these are true aliases and not copies, values are fully interchangeable between the old and new import paths — nothing can drift between them.

**This shim should be removed** once meshkit's `go.mod` pin on `github.com/meshery/schemas` is bumped past v1.3.26 to a version that no longer references the old `meshmodel` path. Flagging here so it doesn't get left behind by accident — happy to open a follow-up PR for the removal once that dependency bump happens.

`go build ./...` and `go test ./...` both pass cleanly on this branch.

## Related

Companion work to meshery/meshery#20632 . Part of the multi-repo capabilities registry rename effort.

Signed commits


models/meshmodel/entity/types.go
```diff
@@ -1,38 +1,39 @@
...
+// It exists solely because github.com/meshery/schemas@v1.3.26 (our currently
+// pinned dependency) has helper files that import this path directly by
+// convention. All types here are pure aliases (type X = Y) of their
+// counterparts in github.com/meshery/meshkit/models/registry/entity, so
+// nothing drifts between the two paths.
+//
+// Deprecated: import github.com/meshery/meshkit/models/registry/entity
+// directly. This shim should be deleted once meshkit's go.mod pin on
+// github.com/meshery/schemas is bumped past v1.3.26 to a version that no
+// longer references the old meshmodel path. Tracked in <ISSUE_LINK>.
```
ishwar170695: Looks like <ISSUE_LINK> is still a placeholder. Could we replace it with the actual tracking issue before merging?

models/meshmodel/entity/types.go
```diff
@@ -1,38 +1,39 @@
+// Package entity is a deprecated compatibility shim for the registry entity
+// package.
+//
+// It exists solely because github.com/meshery/schemas@v1.3.26 (our currently
+// pinned dependency) has helper files that import this path directly by
+// convention. All types here are pure aliases (type X = Y) of their
+// counterparts in github.com/meshery/meshkit/models/registry/entity, so
+// nothing drifts between the two paths.
+//
+// Deprecated: import github.com/meshery/meshkit/models/registry/entity
```
ishwar170695, line 10: Could we also add a TODO here (or reference the tracking issue) so it's easier to remove this compatibility shim once github.com/meshery/schemas no longer depends on the old import path?

ishwar170695:
`AGENTS.md` and `docs/agent-instructions/registration.md` still reference `models/meshmodel/registry`. Since this PR renames that package to `models/registry/manager`, those docs may also need to be updated (either here or in a follow-up PR).

models/meshmodel/entity/types.go
Ayush-kr-giga: you're right that a dangling placeholder isn't great. After discussing with @Maanvi212006, we removed that sentence instead of linking a real issue, since the shim doesn't exist on master yet, so a tracking issue can't meaningfully reference it until this merges. The rest of the deprecation comment still explains the why/when (remove once the schemas pin moves past v1.3.26). Open to opening a proper tracking issue right after merge if that's preferred

models/meshmodel/entity/types.go
Ayush-kr-giga: done

Ayush-kr-giga:
> `AGENTS.md` and `docs/agent-instructions/registration.md` still reference `models/meshmodel/registry`. Since this PR renames that package to `models/registry/manager`, those docs may also need to be updated (either here or in a follow-up PR).

Updated — AGENTS.md and docs/agent-instructions/registration.md now point to models/registry/manager/ instead of the old models/meshmodel/registry/ path. Thanks for catching that

PR #1076, opened 2026-07-28
Propagate DB errors in CategoryFilter.Get

### LINKED ISSUE #1075: CategoryFilter.Get swallows database errors and returns an empty result as success
### Current Behavior

`CategoryFilter.Get` (`models/meshmodel/registry/v1beta1/category_filter.go`) drops the error from the DB query and returns a successful-looking empty result:

```go
err := finder.Find(&catdb).Error
if err != nil {
    return cat, count, int(count), nil   // error dropped, nil returned
}
```

Every sibling filter propagates the error instead. `ModelFilter.Get` (`model_filter.go`):

```go
err := finder.Find(&modelWithCategories).Error
if err != nil {
    return nil, 0, 0, err
}
```

`ConnectionFilter.Get` does the same. `CategoryFilter` is the only outlier.

### Consequence

Any real `Find` failure — a bad `OrderOn` column from a caller-supplied sort param producing a SQL error, or a transient DB error — is reported as **HTTP 200 with an empty category list**, with nothing surfaced to the client and nothing logged. The two `GetMeshmodelCategories*` handlers in meshery also discard the error (`categories, count, _, _ := ...`), so the failure is invisible end to end.

### Desired Behavior

Propagate the error like the sibling filters do (`return nil, 0, 0, err`).

I have the one-line fix ready (GOOS=linux build + vet pass) and will open a PR referencing this issue.

Atishyy27:
Fixes #1075.

### What

`CategoryFilter.Get` dropped the error from `finder.Find` and returned an empty result with a nil error:

```go
err := finder.Find(&catdb).Error
if err != nil {
    return cat, count, int(count), nil   // swallowed
}
```

So a real query failure (e.g. a bad `OrderOn` column from a caller-supplied sort param → SQL error, or a transient DB error) surfaced as an **empty category list reported as success** — nothing returned to the client, nothing logged.

### Change

Return `nil, 0, 0, err`, matching every sibling filter — `ModelFilter.Get` and `ConnectionFilter.Get` both already propagate the error this way. `CategoryFilter` was the only outlier.

### Testing

One-line change; `GOOS=linux go build ./models/meshmodel/registry/...` and `go vet` both pass. Behaviour for the success path is unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected category filtering error handling so failed requests now return the appropriate error instead of appearing successful.
  * Ensured pagination counts are reset when category data cannot be retrieved.

### DIFF
--- a/models/meshmodel/registry/v1beta1/category_filter.go
+++ b/models/meshmodel/registry/v1beta1/category_filter.go
@@ -76,7 +76,7 @@ func (cf *CategoryFilter) Get(db *database.Handler) ([]entity.Entity, int64, int
 
 	err := finder.Find(&catdb).Error
 	if err != nil {
-		return cat, count, int(count), nil
+		return nil, 0, 0, err
 	}
 
 	for _, c := range catdb {


Atishyy27:
hey, noticed CI hasn't run on a few of my PRs here - looks like it's waiting on an "approve and run workflows" since this is my first PR in this repo. same thing on #1078, #1093, and #1094. would appreciate someone unblocking whenever's convenient, thanks!

PR #1109, opened 2026-09-02
Fix goroutine-lifecycle race that can panic on a closed channel

### LINKED ISSUE #1102: registry: "send on closed channel" panic when a model generation times out
`InvokeGenerationFromSheetWithOptions` (`registry/model.go`) has a goroutine-lifecycle race that can panic with `send on closed channel` and crash the whole process when a model's generation exceeds `opts.ModelTimeout`. It is interleaving-dependent, not a crash on every timeout: it fires only when an orphaned worker goroutine reaches its send after the channel has been closed. A run where the worker happens to send before the close just leaks the goroutine instead.

**Mechanism**

For each model an outer goroutine is started and tracked with `wg.Add(1)` (line 899). Inside it, a second inner goroutine (line 917) does the actual generation work (`generator.GetPackage()`, `GenerateComponents()`, the filesystem writes) and sends its result on the unbuffered `spreadsheeetChan` at line 1024.

The outer goroutine then waits on a `select` (line 1038): either the inner goroutine finishes (`<-done`) or the per-model timeout fires (`<-modelCtx.Done()`, line 1055). On timeout the outer goroutine returns and its deferred `wg.Done()` runs, but the inner goroutine is not tracked by `wg` and is not cancelled: neither the work nor the send at 1024 ever checks `modelCtx`. It keeps running.

Once every model has finished or timed out, `wg.Wait()` (line 1065) returns and `close(spreadsheeetChan)` runs (line 1066). An inner goroutine still alive from a timed-out model that later reaches line 1024 sends on the now-closed channel, which panics. A goroutine panic is fatal, so the whole process crashes, and it happens after the run has already logged its completion summary, which makes the crash hard to trace back to its cause.

Even when it does not panic, it is a goroutine and fd leak: the orphaned inner goroutine keeps doing filesystem writes that nothing is tracking.

**Trigger**

Any model whose generation exceeds `ModelTimeout` (default is 5 minutes via `DefaultModelTimeout`). This is realistic under github / artifacthub rate limiting or with large sources, which is the exact situation the timeout was added to bound. `generator.GetPackage()` takes no context and cannot be cancelled, so a slow inner goroutine genuinely outlives the outer one.

**Suggested direction**

The inner goroutine's send has to be safe against both cancellation and the channel already being closed, because the producer can outlive the close (generation is not cancellable). Options: a context-aware send that also cannot panic if the channel is closed, or only closing `spreadsheeetChan` after every producer is accounted for. Happy to open a PR once the preferred approach is clear.

MU5A:
Fixes #1102

`InvokeGenerationFromSheetWithOptions` (`registry/model.go`) starts an inner goroutine per model to do the actual generation work, tracked only through the outer per-model goroutine's own `wg` entry. When a model's generation exceeds `opts.ModelTimeout`, the outer goroutine returns via the `modelCtx.Done()` case and its `wg.Done()` fires — but the inner goroutine is not cancelled (`generator.GetPackage()` takes no context) and keeps running, untracked.

If every other model has since finished or timed out, `wg.Wait()` returns and `close(spreadsheeetChan)` runs while that orphaned inner goroutine is still alive. When it later reaches its send on `spreadsheeetChan`, it panics on a closed channel and crashes the whole process — typically after the run has already logged its completion summary, which makes the crash hard to trace back.

This implements the second option from the issue's own suggested direction: add the inner goroutine to the same `WaitGroup` so `spreadsheeetChan` is only closed once every producer, including orphaned ones from timed-out models, has actually finished with it.

## Test plan
- [x] `go build ./registry/...` and `go vet ./registry/...` clean
- [x] Existing tests in `registry/model_generation_test.go` (timeout behavior, progress tracker, model timeout with/without context) pass
- [x] Traced all early-return paths in the inner goroutine — the added `defer wg.Done()` covers every one of them, including the error-exit paths that existed before this change

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of timed-out model generation requests.
  * Prevented potential processing errors and application crashes when results arrive after a request has finished.
  * Added safer handling for late spreadsheet-generation results, ensuring they are ignored gracefully when processing has already closed.

### DIFF
--- a/registry/model.go
+++ b/registry/model.go
@@ -914,7 +914,19 @@ func InvokeGenerationFromSheetWithOptions(wg *sync.WaitGroup, path string, model
 			// Channel to receive generation result
 			done := make(chan error, 1)
 
+			// Tracked on wg independently of the outer goroutine: if this
+			// model times out, the outer goroutine returns via the
+			// modelCtx.Done() case below and its own wg.Done() fires, but
+			// this inner goroutine is not cancelled and keeps running
+			// (GetPackage takes no context). Without its own wg entry,
+			// wg.Wait() below can return - and close(spreadsheeetChan) can
+			// run - while this goroutine is still alive, so its later send
+			// on spreadsheeetChan panics on the closed channel. Adding it
+			// here keeps the channel open until every producer, including
+			// orphaned ones from timed-out models, is done with it.
+			wg.Add(1)
 			go func() {
+				defer wg.Done()
 				var genErr error
 
 				if utils.ReplaceSpacesAndConvertToLowercase(model.Registrant) == "meshery" {


lekaf974 (approved):
LGTM

emphor11:
@MU5A Could we revisit tying the inner generation goroutine to this `wg`?

The `ModelTimeout` currently only stops waiting in the outer goroutine; the inner worker does not use `modelCtx` and therefore can continue indefinitely. By adding that worker to `wg`, `wg.Wait()` now waits for the timed-out work before closing the channel and returning from `InvokeGenerationFromSheetWithOptions`.

That prevents the closed-channel panic, but it also means a timed-out model can keep the entire generation run blocked indefinitely. Since the timeout exists to bound slow model generation, could we instead separate the producer/channel lifecycle from the function's completion, or otherwise ensure the timed-out worker cannot hold the overall operation open?

MU5A:
@emphor11 you're right, and that's a real regression, not a nitpick. Reverted the WaitGroup change.

Pushed a different fix: instead of making `wg.Wait()` wait for orphaned inner goroutines (which reintroduces the unbounded-block problem you flagged), I wrapped the actual send in a `recover()`-guarded helper (`safeSendSpreadsheetData`). The outer per-model goroutine's `ModelTimeout` bound on the overall run is now exactly what it was before this PR — unchanged. If a model's inner goroutine is still alive after `close(spreadsheeetChan)` and tries to send its late result, that send is now safely discarded instead of panicking. The model was already counted as timed out either way; we just stop that late, orphaned write from taking the whole process down with it.

Added a regression test (`TestSafeSendSpreadsheetDataOnClosedChannel`) exercising the recover path directly. Build/vet/full package tests all pass. Thanks for catching this before merge.

MU5A:
Read both comments carefully, they are pointing at something real, but I want to be precise about what this PR does and does not claim to fix before responding.

Verified against git history: a model that finishes generating just after its own ModelTimeout, but before spreadsheeetChan actually closes, has always been silently published, in the original code before this PR existed too. The original code did an unconditional spreadsheeetChan <- SpreadsheetData{...} with zero gating on timeout state, identical in that respect to the code today. This PR's scope was narrower: stop the crash when a result arrives after the channel has already closed. It never claimed, and does not now claim, to make publication consistent with the per-model timeout outcome.

That inconsistency (logs say TIMEOUT, output still contains the model anyway) is real and worth fixing, but it's a separate, pre-existing design question this PR didn't introduce: should a late-but-valid result be kept (more complete output) or discarded (logs and output agree)? That's a product decision, not just a bug fix, and building it properly needs synchronized per-model timeout state checked at the point of publication, exactly what the second comment describes, which is a real feature addition, not a small tweak.

Rather than fold a design decision into a PR that's already been reviewed once for a narrower fix, I'd rather file this as its own issue with both of you credited for raising it, unless you'd prefer I just implement it here. Which would you rather?

PR #928, opened 2026-03-07
feat: clear $ref fields across entire OpenAPI document

guan404ming:
**Description**

This PR fixes #926

**Notes for Reviewers**

## Why                                                                                                 
  
The `getResolvedManifest` function only cleared `$ref` fields from `components/schemas`, leaving refs in paths, parameters, responses, request bodies, headers, and callbacks unresolved.
                                                                                                       
## How                                                          

- Extract ref-clearing into `clearDocRefs` that traverses paths, all component types, and nested
callbacks
- Add helpers for clearing refs in parameters, headers, request bodies, responses, operations, and
content
- Add tests verifying ref resolution in paths, request bodies, responses, and headers

**[Signed commits](../CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [x] Yes, I signed my commits.

PR #924, opened 2026-03-04
Replace hand-rolled $ref resolver with kin-openapi

guan404ming:
**Description**

## Why                                                                                                                   
   
The custom $ref resolver was fragile and limited; kin-openapi provides a battle-tested OpenAPI parser that resolves   
references automatically.                                       

## How

- Replace the manual CUE-based $ref resolution in getResolvedManifest with openapi3.Loader from kin-openapi
- Add clearSchemaRefs to recursively strip $ref strings so json.Marshal outputs fully inlined schemas
- Add comprehensive tests covering property refs, nested chains, allOf, array items, additionalProperties, and YAML input

**Notes for Reviewers**

**[Signed commits](../CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [ ] Yes, I signed my commits.

PR #1110, opened 2026-09-07
Clear schema refs across the whole document, not just components.schemas

### LINKED ISSUE #926: Increase robustness of manifest handling in model generation.
> medium
> 
> The current implementation of `getResolvedManifest` only clears `$ref` fields from schemas within `components/schemas`. While this is sufficient for the current usage in `GenerateFromOpenAPI`, the function is not fully generic. An OpenAPI document can have `$ref`s in other places like `paths`, `parameters`, `responses`, etc. To make this function more robust and reusable, consider extending it to clear `$ref`s from the entire document. For example, you could iterate through `doc.Paths` and other parts of `doc.Components` and call `clearSchemaRefs` on any `SchemaRef` found. 

 _Originally posted by @gemini-code-assist[bot] in [#924](https://github.com/meshery/meshkit/pull/924/changes#r2881678676)_

MU5A:
Fixes #926

Originally flagged by a Gemini code review comment on #924: `getResolvedManifest` only cleared `$ref`s on `doc.Components.Schemas`. That's sufficient for the current single call site (`GenerateFromOpenAPI`, used for GitHub-hosted registrants), but the function's own job is to fully inline schemas for an arbitrary third-party OpenAPI document, and a `$ref` inside a path operation's parameters, requestBody, response, or response header is exactly as reachable in the final `json.Marshal` output as one in `components.schemas`.

A prior attempt at this (#928) was opened and later closed unanswered ("Is this needed any longer?"). Checked the current code before starting: the gap is still there, and the only real call site (`generators/github/package.go`) is the same model-generation pipeline this repo already ships to production.

## What I changed

Added `clearDocumentSchemaRefs` and its helpers, walking `doc.Paths` (every operation verb per path, plus path-level parameters) and the non-schema parts of `doc.Components` (parameters, requestBodies, responses, headers), reusing the existing `clearSchemaRefs` and its cycle-detection stack.

Deliberately skips any parameter, request body, response, or header reached through a non-empty `$ref` rather than walking into its `Value`. kin-openapi's `*Ref` types all marshal to `{"$ref": ...}` and drop `Value` entirely whenever `Ref` is set, so clearing schema refs nested inside one would never be visible in the output, and mutating that shared `Value` risks corrupting every other reference to the same component.

## Test plan
- [x] Added a regression test exercising a `$ref` inside an inline path operation's `requestBody` and `response`
- [x] Full `utils/component` package tests pass, including the existing real-world Kubernetes OpenAPI spec integration test (`TestGetResolvedManifest_RealKubernetesSpec`)
- [x] `go vet` and `gofmt` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OpenAPI manifest generation so referenced schemas are resolved and inlined across operations, request bodies, responses, parameters, headers, callbacks, and webhooks.
  * Preserved unresolved schema references instead of producing invalid empty references.
  * Prevented errors when unresolved references appear in encoding headers.

* **Tests**
  * Added coverage for schema resolution across callbacks and OpenAPI 3.1 webhooks.
  * Added regression coverage for unresolved references in encoding headers.


suryaff733:
@MU5A, could you please check the suggestions from the CodeRabbit bot?

MU5A:
@suryaff733 done, thanks for the nudge. Addressed both real gaps CodeRabbit flagged:

1. **Callbacks**: op.Callbacks was never walked. A callback is a map of runtime expression to PathItem, one level below the operation, so its request/response schemas were unreachable before. Added clearCallbackSchemaRefs and wired it into both the operation-level and components-level walks.

2. **Encoding headers**: a multipart part's own headers (declared under content.<type>.encoding.<part>.headers) can carry a schema $ref and were not walked either. Fixed that too.

Extending into encoding headers surfaced a real latent bug in clearSchemaRefs itself: it cleared SchemaRef.Ref unconditionally, before checking whether Value was ever resolved. Turns out kin-openapi's loader never resolves a $ref at that specific location, Value stays nil while Ref is set. Clearing Ref there produced an empty, invalid SchemaRef{Ref:"", Value:nil} that panicked inside SchemaRef.MarshalJSON. Fixed clearSchemaRefs to only clear Ref when there's a resolved Value to put in its place. That one specific corner (unresolved refs at that location) still shows up as a $ref rather than fully inlined, actually resolving it ourselves would mean reimplementing $ref resolution, which felt like real scope creep for this PR, so I left it honest rather than silent.

Also added the docstrings CodeRabbit's coverage check wanted, and a regression test for each of the three things above. Full package tests pass.

MU5A:
CodeRabbit's consolidated follow-up named three remaining branches (Callbacks, Webhooks, Encoding headers). Callbacks and Encoding headers were already covered by the previous commit, but it also marked its own comment "Addressed in commit 7a09752" for doc.Webhooks specifically, which was not true, that field was never touched. Checked the current code directly rather than trusting the marker, confirmed the gap, and pushed a fix.

doc.Webhooks (OpenAPI 3.1+) is a map of name to PathItem describing requests the API sends out rather than ones it serves, same shape as doc.Paths, walked the same way now. Added a regression test. Full package tests pass.

PR #1121, opened 2026-09-20
feat(walker): hybrid GitHub Trees crawl with selective blobs and interesting-file ranking

### LINKED ISSUE #1119: feat: Hybrid GitHub Trees crawl + selective blobs + interesting-file ranking
## Summary

Hybrid GitHub crawl for design import: use **Git Trees API + selective blobs + interesting-file ranking**; keep **go-git** for non-GitHub hosts and truncated-tree fallback. Today meshkit `walker` shallow-clones the whole repo (`PlainClone`, depth 1) then filters YAML — costly for large repos; Contents/Trees walkers already exist in places.

Research: https://github.com/layer5labs/meshery-extensions/pull/4366 · meshkit pointer https://github.com/meshery/meshkit/pull/1118

## Acceptance criteria

- [ ] Fix branch/ref wiring on clone options (Branch vs ReferenceName gap)
- [ ] Context timeouts + progress hooks around clone/HTTP walkers
- [ ] GitHub path: resolve ref → recursive Trees → filter → selective blob fetch
- [ ] Auth: wire GitHub App/OAuth token from connection into walkers
- [ ] Truncated tree → fall back to go-git
- [ ] Non-github.com hosts keep go-git
- [ ] Ranked interesting-file list (Helm Chart.yaml, compose, kustomization, design.yml, yaml/yml/json, chart archives) **before** blob download
- [ ] Reuse existing file identification (Helm/K8s/Compose/Kustomize/OCI/design)

## Priority

P2 (can parallelize after CONNECT; unlocks better Canvas/Cloud pickers)

## Related

- Cloud/Extensions import UX consumes this list/crawl API

pontusringblom:
Adds an opt-in hybrid GitHub crawl to `utils/walker` for design import. Today `Git.Walk` shallow-clones the whole repository and then filters it, which is costly on large repos. The new route resolves the ref to a commit, lists the tree once recursively, ranks the entries from metadata alone, and downloads only the blobs that survive — keeping go-git for every case the API cannot answer completely.

The crawl is opt-in via `UseGithubAPI()`. Callers that do not enable it keep cloning as before, with one deliberate exception noted below.

## Acceptance criteria

- **Branch/ref wiring** — `Branch()` was recorded but never reached `CloneOptions`, so a caller setting `Branch("foo")` silently cloned the default branch. An explicitly set branch now becomes `refs/heads/<branch>`; an explicit `ReferenceName` still wins; a caller setting neither still gets the remote's default branch.
- **Context and progress** — `WalkContext`, `ListInterestingFiles` and `FetchCandidates` take a context; `Walk()` delegates with `context.Background()`. `Timeout(d)` bounds a traversal, and `RegisterProgressHook` reports `resolve-ref`, `list-tree`, `rank`, `fetch-blob` and `clone` stages. Both HTTP walkers use `http.NewRequestWithContext`; the clone uses `PlainCloneContext`.
- **Trees route** — resolve ref → recursive Trees → filter and rank → selective blob fetch. The `MaxFileSize` limit is applied against the size the tree reports, so an oversized blob is not requested, and again against the decoded contents so the configured limit holds even when a caller-supplied candidate omits its size.
- **Auth** — `Token()` threads a GitHub App or OAuth token onto the API calls as a bearer token and onto the clone as `x-access-token` basic auth. The token is never logged, never placed in an error message and never reported through a progress hook.
- **Truncated tree → go-git** and **non-github.com hosts → go-git** — the host is read from the configured `BaseURL` rather than assumed, so GitHub Enterprise, GitLab, Bitbucket and `file://` URLs keep the clone route.
- **Ranked interesting-file list before download** — `ListInterestingFiles(ctx)` returns ranked candidates (path, name, SHA, size, inferred kind, score) with no blob fetched, so a picker can be rendered and the selection passed back to `FetchCandidates`.
- **Reuse of existing identification** — ranking uses `core.IaCFileTypes` and the existing extension tables rather than restating literals. Full identification via `files.IdentifyFile` remains the caller's job once contents are in hand.

## Behaviour changes for callers that do not opt in

- The branch/ref fix above applies to every caller: `Branch("master")` against a repository whose default branch is `main` now fails where it previously cloned the default branch and succeeded. That is the fix working as specified.
- Symlink handling on the clone route: a symlink whose target resolves outside the temporary clone, or that cannot be resolved, is passed over instead of being read through. A link resolving inside the clone is followed as before.

## Notes

- Ranking is path-based because it runs before any content exists; `CandidateFile.Kind` is a hint, not an identification.
- The `Github` (Contents API) walker gains `Token`, `WalkContext` and `RegisterProgressHook` alongside test coverage it previously lacked.
- Internal documentation is in `docs/git-walkers.md`, linked from the architecture doc and the agent instruction index.
- New error codes `meshkit-11328`–`11334`.

Fixes #1119

Design research: layer5labs/meshery-extensions#4366, and the merged pointer doc `docs/ux/canvas-first-github-onboarding.md` (#1118).

## Summary by CodeRabbit

- **New Features**
  - Added hybrid GitHub repository crawling with automatic fallback for unsupported or incomplete API responses.
  - Added context cancellation, timeouts, authentication tokens, progress reporting, and configurable branch/reference handling.
  - Added ranked discovery and fetching of likely configuration and design files.
  - Added clearer errors for invalid roots, references, file sizes, and repository access issues.

- **Documentation**
  - Added comprehensive walker documentation, architecture references, and onboarding status updates.

- **Bug Fixes**
  - Improved path handling, symlink containment, size-limit enforcement, cancellation behavior, and secure clone directory permissions.


arjunmehta-git (approved):
Reviewed against the acceptance criteria in #1119. All five checks are green on head `a2dcaf81` (DCO, lint, test, tidy, reviewer), verified per-SHA rather than from the aggregated PR view.

**Acceptance criteria, each confirmed present:**

- Branch/ref wiring fixed — a configured branch now reaches the clone options instead of being silently ignored, and an unset branch resolves the repository's real default branch rather than a literal `master`. Verified against the live API on a repository whose default branch is neither `master` nor `main`.
- Context timeouts and progress hooks on both walkers, with the existing entry points delegating so current callers are unaffected.
- GitHub path resolves the ref, lists the tree recursively in one request, filters, and fetches only surviving blobs.
- Token auth threaded through both the API calls and the clone, never logged and never placed in a URL or error.
- Truncated tree falls back to go-git; non-github.com hosts keep go-git unchanged.
- Ranked candidate listing returned from tree metadata alone, with no blob downloaded.
- Kind inference reuses the existing file-identification taxonomy rather than duplicating it.

**Defects found during review and corrected before merge**, the substantive ones being:

- A ranked listing could mislabel ordinary archives as Helm charts and rank them above real manifests.
- Several paths reported success while delivering nothing — a missing root, a folder-scoped listing dropping nested charts, a selection fetched with no interceptor registered, and an explicitly named file that ranking filtered out. All now fail with a clear reason.
- A credential configured for one forge could be transmitted to another; both new entry points now refuse a non-github.com host.
- An error-propagation change would have failed an entire walk on ordinary repository shapes such as a submodule or symlink; narrowed to cancellation only.
- Symlinked roots given one determinate contract across every entry point, with no silent case remaining.

**Back-compatibility:** the new behaviour is opt-in. Callers that do not opt in keep their existing results, including path semantics and the existing oversize error. This was treated as a hard constraint throughout, since this package is consumed outside the repository.

**Not inspected here:** runtime behaviour against large live repositories, and downstream integration in the consuming products. The test suite covers the walker paths against local stubs rather than the network.

